PR #25, opened 2018-06-08
Installs the v2 version of  grommet-cli and grommet library.

devdil:
npm install grommet points to the older version.To install the v2, we
have to use npm install grommet@next.


alansouzati:
thanks for your contribution. welcome to the team!

PR #28, opened 2018-08-31
Added playground. Removed theme creator.

alansouzati:
Theme creator was nice, but the final output was a json object and users were confused on what to do with that.

I'm replacing the theme creator page with a new feature called playground. It uses the vscode monaco editor, and a live preview. I've added two playgrounds for now: Hello World and Theme.

I will work on add more playgrounds soon. but I feel this is a great starts towards a consistent and pleasant on-boarding experience for grommet folks.


src/screens/Play.js
```diff
@@ -0,0 +1,112 @@
+import React, { Component } from 'react';
+import styled, { css } from 'styled-components';
+import MonacoEditor from 'react-monaco-editor';
+/* eslint-disable import/no-duplicates */
```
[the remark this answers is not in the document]
alansouzati: we need to pass the scope to the LiveProvider.

webpack.config.js
```diff
@@ -4,36 +4,39 @@ const CleanWebpackPlugin = require('clean-webpack-plugin');
...
-            && context.resource.search(/[\\/]node_modules[\\/]/) !== -1
-            && context.resource.search(/[\\/]node_modules[\\/]grommet/) === -1
-          ),
-          name: 'vendors',
-          priority: -10,
-        },
-      },
-    },
-  },
+  // does not work with monaco editor.
```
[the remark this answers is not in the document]
alansouzati: it does not highlight the code :( I could not trace down why it does not work.

webpack.config.js
```diff
@@ -57,6 +32,35 @@ module.exports = {
...
+          test: /[\\/]node_modules[\\/]grommet/,
+          name: 'grommet',
+          priority: -10,
+        },
+      },
+    },
+  };
+} else {
+  baseConfig.plugins.push(new webpack.HotModuleReplacementPlugin());
+  baseConfig.devServer = {
```
ericsoderberghp: I think this is automatically ignored in production.

webpack.config.js
```diff
@@ -57,6 +32,35 @@ module.exports = {
...
+        test: /\.css$/,
+        use: [ 'style-loader', 'css-loader' ]
+      }
     ],
   },
 };
+
+if (env === 'production') {
+  baseConfig.plugins.push(new OfflinePlugin());
+  baseConfig.optimization = {
```
ericsoderberghp: I think this is automatically ignored in development.

package.json
```diff
@@ -14,7 +14,7 @@
     "build": "cross-env NODE_ENV=production webpack --mode production",
     "build-alias": "cross-env USE_ALIAS=true NODE_ENV=production webpack --mode production",
     "release-heroku": "babel-node ./tools/release-heroku.js",
-    "start": "webpack-dev-server --mode development --open",
+    "start": "cross-env NODE_ENV=development webpack-dev-server --mode development --open",
```
ericsoderberghp: Why did we need to add this?

I'm asking because most of webpack 4 is supposed to be smarter about this sort of thing with `mode`.

public/index.html
```diff
@@ -11,6 +11,7 @@
   <meta name="theme-color" content="#865CD6"/>
   <link rel="shortcut icon" type="image/png" href="/img/shortcut-icon.png" />
   <link rel="apple-touch-icon" sizes="196x196" type="image/png" href="/img/mobile-app-icon.png">
+  <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Roboto:300,400,500">
```
ericsoderberghp: Why do we need Roboto?

ericsoderberghp:
From a UX perspective, I find the fact that the "play" item in the top navigation menu behaves differently compared to the other menu items a bit odd. I'm wondering if it shouldn't be a top level navigation item but instead a big button on the page somewhere.

ericsoderberghp:
Do we need to have the Eclipse style thumbnail overview of the text? It seems cluttered.

ericsoderberghp:
It has some issues on mobile. I don't think we should try to show them side-by-side on mobile. I think we should switch back and forth between code and display.

package.json
alansouzati: I dont think the dev server is there yet. Without it,
It is running with empty node env. Here I’m just making sure we are running with that env variable set. I don’t think it hurts. Specially because we are relying on that in the webpack config

webpack.config.js
alansouzati: To show a different theme in the theme playground

alansouzati:
> From a UX perspective, I find the fact that the "play" 

I agree. I'm waiting the latest design from @L0ZZI. We can change where we trigger this (I would not block this PR waiting for that though, given that I feel this PR offers a great feature to help new and existing users onboard).

> Do we need to have the Eclipse style thumbnail overview of the text? It seems cluttered.

This can be removed, I agree. For our examples, we are not expecting a lot of content anyways.

> It has some issues on mobile. I don't think we should try to show them side-by-side on mobile. I think we should switch back and forth between code and display.

I don't think we should support mobile for this. Plus, I don't think Monaco Editor has proper mobile browser support in the first place.

https://github.com/Microsoft/monaco-editor/issues/246

Personally, I don't think people would use the playground on mobile.

Issue #27, opened 2018-08-29
The start reading link leads to a blank page but refreshing shows page correctly


conorhastings:
![grommetv2startreading](https://user-images.githubusercontent.com/8263298/44800056-719d4d80-ab83-11e8-9762-9c0179a88e5a.gif)

spencerkordecki:
@conorhastings I opened a PR for this, should be fixed next time a release is made!

alansouzati:
thanks for the PR @spencerkordecki

PR #31, opened 2018-10-07
update get started button path on home screen

spencerkordecki:
Fixes issue #27 by changing the path on the Getting Started `RoutedButton` on the home page from `path='components'` to `path='/components'`


ericsoderberghp:
Thanks!

Issue #35, opened 2018-10-09
External links not working


mtendotnet:
The external links (github, slack, storybook, twitter) aren't working on the home page. Noticed that the `<Button as="a" href="...">` components are being rendered as `<button>`s instead of `<a>` tags.

```
<button class="StyledButton-sc-323bzc-0 hnsXoe" as="a" aria-label="Go to Github" href="https://github.com/grommet/grommet/tree/NEXT">
  <div class="StyledBox-sc-13pk1d4-0 ksxcgD">GitHub</div>
</button>
```

Updating `styled-components` dependency from `v3` to `v4` fixes this. Will create PR for this.

alansouzati:
thanks martin makes sense. a PR is gladly appreciated!

PR #36, opened 2018-10-09
Update styled-compenents dependency from v3 to v4

mtendotnet:
Update to `syled-components` v4 to fix issue #35.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "react-ga": "^2.5.3",
     "react-live": "^1.11.0",
     "react-monaco-editor": "^0.18.0",
-    "styled-components": "^3.3.0",
+    "styled-components": "^4.0.0-beta.11.1",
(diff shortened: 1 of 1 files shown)


ericsoderberghp:
Thanks for working on this. I suspect yarn.lock also needs to be updated so that Travis will pass.

alansouzati:
to achieve this you can delete `yarn.lock` run `yarn cache clean` and run `yarn install`. this will generate a new yarn.lock file.

The reason for that is because we are using the stable grommet release and yarn does not like that.

For reference:

https://github.com/grommet/grommet-site/blob/master/package.json#L23

and

https://github.com/grommet/grommet/wiki/What-is-grommet-stable-and-how-to-use-it%3F

alansouzati:
I will take care of this after merging.

mtendotnet:
Ah ok, noted for next time. Thanks

PR #79, opened 2019-02-11
Add theme docs - Chart & Heading

ShimiSun:


src/screens/Chart.js
```diff
@@ -70,7 +70,7 @@ export default () => (
       }}
     />
 
-    <Doc name="calcs" nav={false} desc={descCalcs} />
+    <Doc name="calcs" nav={false} desc={descCalcs} themeDoc={themeDoc} />
```
ericsoderberghp: This looks like it under “calcs” instead of “Chart”

PR #81, opened 2019-02-13
Update Search.js fix .flat() function undefined error

Jumblemuddle:
Fixes this error in browser that don't support `.flat()`:
scrot

Source: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat#reduce_and_concat


ericsoderberghp:
Thanks for your contribution!

PR #91, opened 2019-03-02
Heading - improve color example

ShimiSun:
https://github.com/grommet/grommet/issues/2851

### DIFF
--- a/src/screens/Heading.js
+++ b/src/screens/Heading.js
@@ -18,7 +18,13 @@ export default () => (
       syntaxes={{
         ...genericSyntaxes,
         a11yTitle: 'Open Drop',
-        color: 'neutral-1',
+        color: [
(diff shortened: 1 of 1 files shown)


src/screens/Heading.js
```diff
@@ -18,7 +18,13 @@ export default () => (
       syntaxes={{
         ...genericSyntaxes,
         a11yTitle: 'Open Drop',
-        color: 'neutral-1',
+        color: [
+          'neutral-1',
+          {
+            dark: 'dark-1',
```
ericsoderberghp: I think the example colors should be reversed. If the background color is 'dark', the text color should be light.

ShimiSun:
If we like the additional color example, I can add it on some more relevant components.

PR #107, opened 2019-05-03
fix: add scroll position reset to route changes

### LINKED ISSUE #106: Components button at bottom on homepage does not reset scroll position
The components button at bottom on homepage takes user to correct location `/components` and does not reset the scroll position to 0,0. A new user thus lands on Utilities section of components first and may access the section that is not most important. If the scroll position were to reset to 0, 0 it takes user to a more relevant section - Layout.

crashuniverse:
Currently when route changes scroll position doesn't
reset. As a result when a user reaches to bottom of
a page & changes route, the most important part of
next route is not immediately visible. With scroll
reset on route changes, user sees the top of next route
first.

Fixes #106

### DIFF
--- a/src/Router.js
+++ b/src/Router.js
@@ -43,6 +43,7 @@ export class Router extends Component {
           `${nextPath}${search || ''}`,
         );
         this.setState({ path: nextPath });
+        window.scrollTo(0, 0);
       }
     }
   };


ericsoderberghp:
Thanks for your contribution! Sorry for the delay in merging it.

crashuniverse:
Thank you so much @ericsoderberghp

PR #120, opened 2019-09-23
Responsive context docs

taysea:
Responsive Context had theme docs that were not displaying. This PR allows any theme docs to be displayed.

Local testing has been done to see that this change does display theme docs.


ShimiSun:
Good catch!!! 💟

Issue #122, opened 2019-09-25
Accordion/Chart have multiple 'components' listed on their page


taysea:
When I scroll down past the initial set of prop/theme definitions for Chart, I see a second section also labelled "Chart" but with a different description and no additional props.

When I do the same of Accordion, I see a section another section titled Accordion but this refers to an accordion panel. I didn't even know if was there because I wouldn't expect 2 'components' on one page. Would it make sense for Accordion Panel to have its own page? Or if not, how can we indicate that there is more info about accordion panel because I came across this purely on accident when I did a big scroll.

ShimiSun:
Yea, good catch.
The Heading below the Chart should be Calc. and below Accordion it's Accordion Panel, its a recent regression, mind taking a look?

PR #123, opened 2019-09-26
Update Accordion Panel and Calc Titles

taysea:
This PR resolves the issue noted in Issue #122. Chart has a second item that needed to be labeled "Calc" and Accordion has a second item that needed to be labeled "Accordion Panel".


ShimiSun:
If we could also add a fix for `Tabs` as part of this PR that will be great... :)

PR #127, opened 2019-10-05
Refactored ResponsiveContext screen to use hooks

sofiageo:
Updated screen to use hooks.


ShimiSun:
Thank you for your contribution!

PR #129, opened 2019-11-04
Adjust Search Button behavior

taysea:
This PR changes how the search button behaves when a user clicks on it. When a user clicks on the search button, a TextInput expands and the drop suggestions appear. When the search drop is open, the 'components' button disappears. This helps accommodate the layout for mobile. This change is a proposed adjustment to the current behavior, and I am open to talking more about if this is a better solution or if more adjustments should be made.


src/components/Search.js
```diff
@@ -10,12 +11,18 @@ const allSuggestions = structure.sections
   .reduce((acc, val) => acc.concat(val), [])
   .sort();
 
-export default () => {
+const SearchButton = ({ setSearchOpen, searchOpen }) => {
```
ericsoderberghp: Wondering if we should keep the name `Search` to align with the file name, and because it's a bit more than a `Button`?

ericsoderberghp: I think from this components perspective, these could just be `open` and `setOpen`.

taysea: The issue I was facing here was that I was getting an error that 'Search' is already defined because there is a grommet-icon called Search that I'm importing. I agree that there could be a more encompassing name than `SearchButton` since it is more than the button.

taysea: I agree. Let me change that.

PR #136, opened 2019-12-12
MonacoEditor tooltips get cut off inside their Box 

Jonathan-Mckenzie:
MonacoEditor provides tooltip explanations for hovering over various components and attributes in the code.
The tooltip gets cut off from the top due to the hidden overflow attribute applied to the parent Box.

Removing the overflow attribute does not appear to effect anything else inside the box.

Screenshot at Dec 11 18-42-44

Screenshot at Dec 11 18-42-15


ericsoderberghp:
Thanks for your contribution!

PR #140, opened 2019-12-17
Add static render build step

karatechops:
This PR adds static HTML file rendering as a secondary build step. I've attempted to leave the original code base as untouched as possible and limited myself to adding small checks for `window` wherever necessary. The Monaco code editor proved to be quite difficult and it has been refactored to a re-usable component which only renders when `window` is available. It seems that this library is quite troublesome and it may be worth the effort to refactor it out in the future.

The static build is generated by crawling the app for relative links. If a page does need to live outside the navigable app it will need to be added to the static build code. Alternatively, the stand-alone page could just not be rendered statically.  

As Grommet doc's routing needs are not very in-depth the static page builder does not account for nested links such as `/components/heading`, the static build code assumes routing will remain one page deep. It would not be much effort to build support for this when the need arises.

I've also [fixed a race condition](https://github.com/karatechops/grommet-site/commit/6da8fab05ee2024aabed5f76ad0c12dfb1fdc96b) when the app is navigating to a deep link.

The static site can be previewed here -> https://grommet-site-static.netlify.com/


ShimiSun:
Follow up with google-web-master and Google-Analytics.

PR #167, opened 2020-05-01
Change margin to include 'none'

ShimiSun:
The Heading `margin` doesn't show the 'none' option. This PR attempts to fix it.

### DIFF
--- a/src/utils/props.js
+++ b/src/utils/props.js
@@ -1,9 +1,9 @@
-
 export const genericSyntaxes = {
   a11yTitle: 'a user friendly label for screen readers',
   alignSelf: ['start', 'center', 'end', 'stretch'],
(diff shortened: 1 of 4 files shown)


src/utils/props.js
```diff
@@ -1,9 +1,9 @@
-
 export const genericSyntaxes = {
   a11yTitle: 'a user friendly label for screen readers',
   alignSelf: ['start', 'center', 'end', 'stretch'],
   gridArea: 'a parent grid area name',
   margin: [
+    'none',
```
ericsoderberghp, line 6: 'none' only applies to Heading and Paragraph. The other places we accept margin don't use this value.

ShimiSun: Good catch! I'll take another look

PR #175, opened 2020-07-07
add to checkboxgroup and radiobuttongroup docs

britt6612:
Closes issue #4183 in grommet 

`RadioButtonGroup` and `CheckBoxGroup` are both made from boxes these both have a default gap that is being used. Users can change this so documentation was needed here to let users know that the gap was being set but can be changed. 

https://github.com/grommet/grommet/blob/master/src/js/components/CheckBoxGroup/CheckBoxGroup.js#L12

This is the case for both components.
Preview of what page will look like: 

Screen Shot 2020-07-07 at 11 02 53 AM


ShimiSun:
Looks great, thank you!! 💟

PR #176, opened 2020-07-13
Adjust onChange example

taysea:
onChange does not include the date, it only includes the time portion. Updating to reflect the functionality.

Closes https://github.com/grommet/grommet/issues/4251


src/screens/Clock.js
```diff
@@ -18,7 +18,7 @@ export default () => (
       syntaxes={{
         ...genericSyntaxes,
         hourLimit: [12, 24, '12', '24'],
-        onChange: "('2018-10-23T10:37:46') => {}",
+        onChange: "('T10:37:46') => {}",
```
ShimiSun, line 21: Can you say a few words why this fixes the issue? I forgot the syntax.

taysea: When I did console log of the onChange function, it never printed the date. It only prints the time: https://github.com/grommet/grommet/blob/000ff1d09d3e8f27ec6a416ecb4a98a8775f5f10/src/js/components/Clock/Clock.js#L131-L138

PR #178, opened 2020-08-20
Improved the Form errors and infos props doc

MikeKingdom:
The errors and infos props for Form allow an object with keys for fields
in the Form. The value for the keys can be a string message or a JSX
node. But the way it was documented is confusing and makes it seem like
you should pass an array for each key's value.

This change makes the doc a little more clear and matches how FormField
documents the same properties.

Fix for #4101 in grommet/grommet


ShimiSun:
closes https://github.com/grommet/grommet/issues/4101

Issue #193, opened 2021-01-04
Documentation incorrectly describes how to use the `icons` property


LiveFreeAndRoam:
This applies to both the Accordion and Calendar documenation.

As discussed in [Slack #general](https://grommet.slack.com/archives/C09QQEG69/p1609715837442700).

The documentation incorrectly describes how to use the `icons` property.  E.g.

**accordion.icons.collapse**
The icon to use when the panel is expanded. `   <UpIcon />`

However, using the icon this way leads to the JS Error:
```
Error
Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: object.

Check the render method of `Accordion`
```

Instead, it should be `UpIcon`, not `<UpIcon />`.

LiveFreeAndRoam:
See PR #194.

PR #194, opened 2021-01-04
icons must be functions, not objects

LiveFreeAndRoam:
The `icons` property accepts functions.  Using `<IconName />`, leads
to the JS Error: "Element type is invalid".  Instead, use `IconName`.

Fixes issue #193.


ShimiSun:
Thank you for your contribution, Matt!

Issue #197, opened 2021-01-11
Grommet Docs is Broken 


MikeWalrath:
Notified by https://twitter.com/BaseDefined/status/1348482693093339136

Users are unable to search and the code examples are not working.

ericsoderberghp:
![Screen Shot 2021-01-11 at 9 01 40 AM](https://user-images.githubusercontent.com/11637956/104213870-9fe5f580-53eb-11eb-8e08-0a83cfb30473.png)

might be related as well

resynth1943:
Ahh, so that's why it's been acting up :thinking: good find, thanks! Will subscribe myself to this issue -- if any help is needed, please reach out :-)

resynth1943:
Is it possible that [this](https://github.com/grommet/grommet-site/commit/7147e5b9429245674767fdd6f5428aa8e6404629) broke the app? I remember it working fine before that. I'll test this tomorrow.

PR #199, opened 2021-01-12
Update build script to account for grommet minified file

halocline:
Modifying build script to account for minified grommet.js now resulting in its own chunk when bundled.

Closes https://github.com/grommet/grommet-site/issues/197

Need to verify https://github.com/grommet/grommet-site/issues/197#issuecomment-758088397

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,3 +1,3 @@
 [build]
-  command = "yarn build"
+  command = "yarn build-static"
   publish = "dist"
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,3 +1,3 @@
 [build]
-  command = "yarn build-static"
+  command = "yarn build"
   publish = "dist"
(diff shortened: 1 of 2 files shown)


ShimiSun:
Thank you for the catch and the fix!
I think for now this fix should be sufficient to put the site back in order, the only thing that worries me is what happens if tomorrow a new package will decide to 'leave' the vendors, and we'll be chasing the same ghosts again, should we spend some time to understand the root cause it happened to grommet and hopefully make sure it wouldn't happen any time soon again for other packages? thoughts?

halocline:
> Thank you for the catch and the fix!
> I think for now this fix should be sufficient to put the site back in order, the only thing that worries me is what happens if tomorrow a new package will decide to 'leave' the vendors, and we'll be chasing the same ghosts again, should we spend some time to understand the root cause it happened to grommet and hopefully make sure it wouldn't happen any time soon again for other packages? thoughts?

Yes, we should dig deeper into the webpack config and better understand our current settings and options. I can create and issue to track and provide background context.

PR #207, opened 2021-02-14
Fix markdown processing

tromgy:
The existing functionality for markdown cleanup was incorrect. It only replaced the _first_ occurrences of &lt; and &gt;. At the same time it did not exclude `code` parts of the markdown content from this processing.

Here's an example of [incorrect output](https://v2.grommet.io/menu) it created:

image

and here's how it would look with this change:

image

- also update yarn.lock for the new grommet so it would install right away without additional steps.


ShimiSun (approved):
Thank you for your contribution

jcfilben:
Closing since this code is no longer relevant after docs were changed to the new format.

PR #213, opened 2021-04-19
Improve Lighthouse Performance with async script

### LINKED ISSUE #211: Fix Google mobile warnings
Fix the mobile warnings identified by Google.

https://search.google.com/search-console/core-web-vitals/summary?resource_id=sc-domain%3Agrommet.io&device=2

taysea:
Adds `async` to script tags that are loaded in the HTML to allow the page to continue parsing while the script is being executed. Additionally, optimized the MonacoWebpack plugin to only import `Javascript` as a language as opposed to the 20+ that are supported by default (minimizes bundle size).

Resource: https://www.w3schools.com/tags/att_script_async.asp

BEFORE:
Screen Shot 2021-04-19 at 3 15 32 PM

AFTER:
Screen Shot 2021-04-19 at 3 10 20 PM

There is clearly still areas to improve re: Speed Index.

The change. in this PR,showed the most prominent improvements to performance in comparison to other methods tried which included the following and were only ever able to reduce the "Time to Interactive" score to 18s:
- Separating `monaco-editor` into its own bundle
- Separating `grommet-icons` into its own bundle
- Lazy-loading Monaco Code Editor component on homepage

Closes #211


tools/renderHtml.js
```diff
@@ -37,6 +37,8 @@ export default ({ body, styles, scripts, helmet }) => `<!DOCTYPE html>
     </noscript>
     ${scripts &&
       scripts.length &&
-      scripts.map(scriptSrc => `<script src="${scriptSrc}"></script>`).join('')}
+      scripts
+        .map(scriptSrc => `<script src="${scriptSrc}" async></script>`)
```
halocline, line 41: 🎉  Good find!

halocline (approved):
Good finds! Looks good to me.

ericsoderberghp (approved):

ShimiSun (commented):
Looking good!! Thanks for working on this @taysea 💟

PR #219, opened 2021-06-15
Add a link to Github repo in the header

CopyJosh:
# Problem

Once you leave the homepage, and even there, it is not so easy to get to the github repo from the website. On every documentation page, there is a link to Storybook and the Codesandox, but getting to the source code is more challenging than it should be. This PR adds a Github link to the header.

Adding the icon to the header was not my initial solution. I thought it would be better to add a "Repository" anchor that would navigate directly to each component's source code in the same way the Storybook and Codesandbox links use the  `getAvailableAtBadge` function. Unfortunately, the URLs are not as clean to generate across all the components and utilities, but maybe this is something that could be given more thought in the future.

# Solution

Screen Shot 2021-06-15 at 10 52 28 AM

### DIFF
--- a/src/components/Nav.js
+++ b/src/components/Nav.js
@@ -1,6 +1,6 @@
 import React from 'react';
-import { Box, Text, ResponsiveContext } from 'grommet';
-import { Grommet as GrommetIcon } from 'grommet-icons';
+import { Box, Button, Text, ResponsiveContext } from 'grommet';
+import { Github, Grommet as GrommetIcon } from 'grommet-icons';
 import RoutedAnchor from './RoutedAnchor';
 import RoutedButton from './RoutedButton';
 import Search from './Search';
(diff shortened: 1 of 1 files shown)


ShimiSun:
As I definitely see a value in having a quick Github link access, I feel like adding this interactive element to the header would be duplication to the one that already exists on the footer, what do you think?

ericsoderberghp:
Since the problem statement mentions getting to GitHub from component pages, what about adding a "GitHub" link next to the Storybook and CodeSandbox links and have it point to the appropriate directory for the component? For example, the Meter component's link would take the caller to https://github.com/grommet/grommet/tree/master/src/js/components/Meter

CopyJosh:
> As I definitely see a value in having a quick Github link access, I feel like adding this interactive element to the header would be duplication to the one that already exists on the footer, what do you think?

I agree it is a duplication, part of the problem is the footer isn't rendered on the component pages. Perhaps it could be?

> what about adding a "GitHub" link next to the Storybook and CodeSandbox links and have it point to the appropriate directory for the component?

I think that would be ideal, and probably a little more effort. I tried adding an object to this array, https://github.com/grommet/grommet/blob/master/src/js/utils/mixins.js#L48 but quickly realized setting the url to something like `https://github.com/grommet/grommet/tree/master/src/js/components/${availabelAt}` doesn't work for every component in the library

ShimiSun:
I'm supportive of adding a footer to the component page that will include the github link, slack link, and any other resources

ShimiSun (commented):
I'm supportive of adding a footer to the component page that will include the github link, slack link, and any other resources.

ericsoderberghp:
How about a deep link to the code for the component next to the Storybook and Codesandbox links?

PR #222, opened 2021-06-23
moved Carousel documentation from Grommet to be contained within site

marcioflaviof:


jcfilben:
Thanks for filing this PR, the direction for the documentation on the grommet site has changed so I am closing this PR and replacing it with #232

PR #224, opened 2021-07-02
New doc format for Anchor component

jcfilben:
The Anchor documentation is changed to use the following format:
```javascript
<Property name="">
  <Description>...</Description>
  <PropertyValue>
    <Description>...</Description>
    <Example defaultValue>...</Example>
  </PropertyValue>
</Property>
```
Each prop can have multiple examples and multiple descriptions.

New files were added to allow for this this change: `ComponentDoc`, `Description`, `PropertyValue`, `PropExample`, `Properties`, `Property`, `ThemeDoc`, and `genericPropExamples`

If all components switch to use the new format the following files can be removed: `Doc`, `Example`, `Prop`, `Props`, `Syntax`, `Value`, `ThemeProps`, `utils/props.js`

The "react/no-unescaped-entities" eslint rule is disabled to allow the use of " and ' in the Description and Example sections

related to https://github.com/grommet/grommet/issues/5353


ericsoderberghp:
Heading in a good direction. I am hoping we can have enough structure to allow splitting up each value type to having it's own example and description. Here's a quick visual to get the idea across: https://designer.grommet.io/?id=grommet-site-props-eric-soderberg-hpe-com
Could that be achieved with the new components added herein?

jcfilben:
The current behavior in this PR is that you can have multiple Descriptions and Examples per property. However, right now there isn't a way to link a specific example with a description. Instead the Descriptions and Examples will display in the order they are listed. If a description takes up multiple lines it will not push the next example down so description example pairs are not always aligned together.

Perhaps I could change the Descriptions and Examples to be more in a grid style where each Description and Example had a row in the grid. This way multiline descriptions or examples would each be aligned with eachother.

In the example you sent, the name of the property has a description directly to the right of it. Currently the name of the property is on its own line and descriptions and examples are below it. Should I change this so that there can be a description on the same line as the name of the prop?

Also the current documentation has the descriptions on the left and examples on the right. If we allow a description to be next to the name of the property as described above, should all the descriptions be moved to the right and the examples on the left?

ericsoderberghp:
I think we should come up with a semantic structuring of the components that would allow adjusting the layout and styling as we go. Perhaps something like:
```
<Property name description>
  <PropertyValue example description />
</Property>
```
or perhaps
```
<Property name>
  <Description>...</Description>
  <PropertyValue>
    <Example>...</Example>
    <Description>...</Description>
</Property>
```

jcfilben:
If you have `<Property name description>` is the description going to the right of the name? And follow up question: if so, should all descriptions be on the right side instead of left? Or should the property description go under the name and descriptions stay on the left side?

ericsoderberghp:
I was looking for a component structure that would allow iterating on that as we go. Currently, I'm inclined towards putting the descriptions on the right, as that might make it easier to scan for what is needed on the left. But, I'd be curious to hear your thoughts as well. Maybe I'll put together a few different ideas in the grommet designer and send those out. It can be helpful to have a visual to refer to.

jcfilben:
I'm fine with having the descriptions on the right, one thing I was thinking about though is if we are making changes component by component, some pages that use the old style of documentation will have the descriptions on the left whereas the components using the new style of documentation will have the descriptions on the right which may be confusing to the user when switching between documentation pages. Maybe we could make the left/right change after all the docs pages are using the new style of documenting since the left/right change won't require any changes to the specific component documentation files.

ericsoderberghp:
That's fine with me. And, a good case for getting the semantic structure right, allowing for that flexibility later.

ericsoderberghp:
Nice!

src/utils/themeDocUtils.js
```diff
@@ -1,149 +1,252 @@
...
+  description: PropTypes.node.isRequired,
+};
+
+export const FocusStyle = () => (
+  <>
+    <Property name="global.focus.border.color">
+      <Description>
+        The border color of the component when in focus.
+      </Description>
+      <PropertyValue>
```
taysea, line 106: Placing the comment here, but it applies to all the examples -- is there any way we can give more of a visual differentiation between the "description" portion of the docs and the "types". For me, it wasn't immediately clear that the second bit of things was telling me what types were accepted. Maybe it would be helpful to have a small heading of "accepted types" or more spacing after the description, etc.

src/components/Doc/ComponentDoc.js
```diff
@@ -0,0 +1,149 @@
...
+import { DomProps } from './DomProps';
+
+export const ComponentDoc = ({
+  align = 'center',
+  availableAt,
+  children,
+  code,
+  description,
+  intrinsicElement,
+  isA,
```
taysea, line 17: can we add a comment around what `isA` means? it was a little confusing of a variable name for me

taysea (commented):
I'm liking this approach and the ability to pair examples more tightly with prop types -- I understand we will be able to iterate on the visual layout more in the future, but just dropped one comment that might be a simple change for this first pass.

src/utils/themeDocUtils.js
jcfilben: In latest push I added additional spacing under the general description

src/components/Doc/ComponentDoc.js
jcfilben: Yes, I added a comment explaining isA in latest push

taysea (commented):
I'm still finding it difficult to discern what is the prop type vs what are potential values. Wondering if we explore using the "code" font as a way to have prop types more clear and then a small description that follows?

Very quick mock up idea, but maybe would need to adjust font sizing

Screen Shot 2021-07-07 at 10 52 07 AM

jcfilben:
I added the prop types above the descriptions, displayed in the code font like shown in the example above. I was having a hard time deciding between medium and large font size, medium looked small but large was a big jump up, so I went with medium and made the prop types bold

taysea:
> I added the prop types above the descriptions, displayed in the code font like shown in the example above. I was having a hard time deciding between medium and large font size, medium looked small but large was a big jump up, so I went with medium and made the weight bold

I think medium and bold is okay for text treatment. I think the overall prop description should always come directly after the prop name. It seems like at times it's coming after the prop type with is a little confusing.

Then, maybe the descriptive text for a prop type should be a smaller text size?

jcfilben:
> I think the overall prop description should always come directly after the prop name. It seems like at times it's coming after the prop type with is a little confusing.

Agreed, in the latest push I fixed this so the general description always appears directly under the prop name and then the prop types and specific definitions under that.

For now I left the descriptions for the prop type text the same size, I tried switching it to small but it looks quite small and was worried it may be hard to read? Let me know what you think, if we should leave the text as medium or switch to small. Here is what small looks like:
Screen Shot 2021-07-08 at 3 16 51 PM

jcfilben:
Summary of discussion with Taylor: we decided to leave the prop type description text size as medium because the small size looked too small. To create more distinction between prop types we decided to add more spacing between them. These changes are reflected in push from Thursday

taysea (commented):
All looks good. The only thing is that some descriptions end with a period and others don't. For consistency, I think we should always end in period or never end in one. Recommending we end in a period unless there are strong opinions otherwise.

jcfilben:
I agree that there should be periods at the end of each description. I fixed this in the latest push.

ericsoderberghp (approved):

PR #228, opened 2021-07-15
Updated Box doc to new format

jcfilben:
Box doc uses the new format for documentation

Additional Changes:
`PropExample.js` - added `<pre>` tag so that examples will maintain indentation and line breaks but will also wrap when the screen is resized. This allowed for <Markdown> tags to be removed in `Accordion.js` and `genericPropExamples.js` because they were only being used to maintain indentation in examples.

`PropOptions.js` - added new file to handle example with "where .. could be" (See Box `background`)

`genericPropExamples.js` - made sure each generic example was wrapped in <PropertyValue>, this caused changes in `Accordion.js` and `Anchor.js`

### DIFF
--- a/src/screens/Box.js
+++ b/src/screens/Box.js
@@ -97,9 +97,7 @@ export default () => (
 
         <Property name="margin">
           <Description>
-            The amount of margin around the component. An object can be
-            specified to distinguish horizontal margin, vertical margin, and
-            margin on a particular side.
+            The amount of margin around the component.
           </Description>
           <GenericMargin />
         </Property>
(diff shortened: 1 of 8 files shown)


src/screens/Box.js
```diff
@@ -21,237 +70,637 @@ export default () => (
...
+          </Description>
+          <GenericGridArea />
+        </Property>
+
+        <Property name="margin">
+          <Description>
+            The amount of margin around the component. An object can be
+            specified to distinguish horizontal margin, vertical margin, and
+            margin on a particular side.
+          </Description>
```
jcfilben, line 103:
```suggestion
          <Description>
            The amount of margin around the component.
          </Description>
```

taysea (approved):

ericsoderberghp (approved):

PR #296, opened 2021-08-13
Fix prop anchor links to scroll to location

### LINKED ISSUE #294: Doc links to specific props not scrolling down the page
A link that references a specific prop on a component page should scroll down to that section when you open it, but instead the link opens, goes to the specific prop section and then takes you back to the top of the page.
Example: https://v2.grommet.io/list#onOrder

jcfilben:
Fixes doc links that reference a specific prop on a component page to scroll down to that section when you navigate to the link.
Closes #294

### DIFF
--- a/src/components/Page.js
+++ b/src/components/Page.js
@@ -1,16 +1,37 @@
-import React from 'react';
+import React, { Component } from 'react';
 import { Box } from 'grommet';
 import Nav from './Nav';
 
-const Page = ({ children, background }) => (
-  <Box pad="large" background={background}>
-    <Box>
-      <Nav />
-      <Box margin={{ top: 'large' }}>{children}</Box>
-    </Box>
-  </Box>
-);
+export default class Page extends Component {
+  componentDidMount() {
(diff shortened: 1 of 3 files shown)


src/components/Page.js
```diff
@@ -1,16 +1,37 @@
...
 
-Page.propTypes = Box.propTypes;
+  componentDidUpdate() {
+    this.scrollToSection();
+  }
+
+  scrollToSection = () => {
+    const name = window.location.hash.split('#')[1];
+    if (name) {
+      document.getElementById(name).scrollIntoView({
```
ericsoderberghp, line 17: Doesn't the browser already do this?
https://stackoverflow.com/questions/19299220/is-there-an-event-that-fires-when-a-browser-scrolls-to-a-named-anchor

jcfilben: The browser supplied event doesn't fire when you go to a url with the hash already set. It only works if you are already on the same page

ericsoderberghp (approved):

PR #298, opened 2021-08-17
Feat/notifications: notification content

amandadupell:
- added main notification documentation for props and theming


src/screens/Notification.js
```diff
@@ -0,0 +1,314 @@
...
+          label: 'Storybook',
+        },
+        {
+          url: 'https://codesandbox.io/s/github',
+          badge: 'https://codesandbox.io/',
+          label: 'CodeSandbox',
+        },
+      ]}
+      description="A Notification"
+      code="// <Notification />"
```
jcfilben, line 33: Right now the code example is causing a syntax error. Maybe change it to {`// <Notification message="..." /> `} and then add more and uncomment once Notification has been added to Grommet.

src/screens/Notification.js
```diff
@@ -0,0 +1,314 @@
...
+
+        <Property name="toast">
+          <Description>
+            A boolean flag that indicates whether or not the Notification is a
+            Toast Notification.
+          </Description>
+          <PropertyValue type="boolean">
+            <Example defaultValue>"false"</Example>
+            <Example>"true"</Example>
+          </PropertyValue>
```
jcfilben, line 75: true and false shouldn't have quotes around them. Also if you prefer you can just use `<GenericBoolFalse />` here, imported from '../utils/genericPropExamples'.

src/screens/Notification.js
```diff
@@ -0,0 +1,314 @@
...
+                        elevation: 'medium',
+                        round: 'xsmall',
+                        width: 'medium',
+                        pad: { horizontal: 'small', vertical: 'xsmall' },
+                        background: {
+                            color: 'background-front',
+                        },
+                    }
+                `}
+            </Example>
```
jcfilben, line 111: When the example is multiple lines, it should be aligned like this to line up with the other examples in the docs. Otherwise there is extra spacing.
```suggestion
            <Example defaultValue>
              {`
{
  elevation: 'medium',
  round: 'xsmall',
  width: 'medium',
  pad: { horizontal: 'small', vertical: 'xsmall' },
  background: {
    color: 'background-front',
  }
}
                `}
            </Example>
```

src/screens/Notification.js
```diff
@@ -0,0 +1,314 @@
...
+        <Property name="notification.iconContainer">
+          <Description>Any valid Box prop for the icon container.</Description>
+          <PropertyValue type="object">
+            <Example defaultValue>
+              {`
+                    {
+                        pad: { right: 'small' },
+                    }
+                `}
+            </Example>
```
jcfilben, line 124: Same alignment comment as above

src/screens/Notification.js
```diff
@@ -0,0 +1,314 @@
...
+        <Property name="notification.textContainer">
+          <Description>Any valid Box prop for the text container.</Description>
+          <PropertyValue type="object">
+            <Example defaultValue>
+              {`
+                    {
+                        gap: 'medium',
+                    },
+                `}
+            </Example>
```
jcfilben, line 137: Same alignment comment as above and remove trailing comma

src/screens/Notification.js
```diff
@@ -0,0 +1,314 @@
...
+              {`
+                    {
+                        weight: 'bold',
+                        color: {
+                            light: 'black',
+                            dark: 'white',
+                        },
+                    },
+                `}
+            </Example>
```
jcfilben, line 158: Same alignment comment as above and remove trailing comma

src/screens/Notification.js
```diff
@@ -0,0 +1,314 @@
...
+            <Example defaultValue>
+              {`
+                    {
+                        color: {
+                            light: 'black',
+                            dark: 'white',
+                        },
+                    },
+                `}
+            </Example>
```
jcfilben, line 175: Same alignment comment as above, and remove trailing comma

src/screens/Notification.js
```diff
@@ -0,0 +1,314 @@
...
+          <Description>The color of the close button icon.</Description>
+          <GenericColor />
+          <PropertyValue>
+            <Example defaultValue>
+              {`
+                    color: {
+                        light: 'black',
+                        dark: 'white',
+                    },`}
+            </Example>
```
jcfilben, line 200: Same alignment comment as above, also remove trailing comma and remove `color:`. In this example you don't need both `{ dark: "string", light: "string" }` and `{ light: 'black', dark: 'white' }` just show the default value for the object example.

src/screens/Notification.js
```diff
@@ -0,0 +1,314 @@
...
+
+        <Property name="notification.unknown.color">
+          <Description>The color of the close button icon.</Description>
+          <GenericColor />
+          <PropertyValue>
+            <Example defaultValue>
+              {`
+                    "status-unknown"
+                `}
+            </Example>
```
jcfilben, line 225: Same alignment comment as above. Show "status-unknown" as the string example instead of "brand". Also PropertyValue should have a type prop.

src/screens/Notification.js
```diff
@@ -0,0 +1,314 @@
...
+
+        <Property name="notification.normal.color">
+          <Description>The color of the close button icon.</Description>
+          <GenericColor />
+          <PropertyValue>
+            <Example defaultValue>
+              {`
+                    "status-ok"
+                `}
+            </Example>
```
jcfilben, line 250: Same alignment comment as above. Show "status-ok" as the string example instead of "brand". Also PropertyValue should have a type prop.

src/screens/Notification.js
```diff
@@ -0,0 +1,314 @@
...
+
+        <Property name="notification.warning.color">
+          <Description>The color of the close button icon.</Description>
+          <GenericColor />
+          <PropertyValue>
+            <Example defaultValue>
+              {`
+                    "status-warning"
+                `}
+            </Example>
```
jcfilben, line 275: Same alignment comment as above. Show "status-warning" as the string example instead of "brand". "status-warning" doesn't need to be wrapped in {``}. Only examples that take up multiple lines or have special characters need to be wrapped with {``}. Also PropertyValue should have a type prop.

src/screens/Notification.js
```diff
@@ -0,0 +1,314 @@
...
+
+        <Property name="notification.critical.color">
+          <Description>The color of the close button icon.</Description>
+          <GenericColor />
+          <PropertyValue>
+            <Example defaultValue>
+              {`
+                    "status-critical"
+                `}
+            </Example>
```
jcfilben, line 300: Same alignment comment as above. Show "status-critical" as the string example instead of "brand". "status-warning" doesn't need to be wrapped in {}. Only examples that take up multiple lines or have special characters need to be wrapped with {}. Also PropertyValue should have a type prop.

jcfilben (commented):
Nice start here, I let some comments below

src/screens/Notification.js
```diff
@@ -0,0 +1,326 @@
...
+        </Property>
+
+        <Property name="notification.textContainer">
+          <Description>Any valid Box prop for the text container.</Description>
+          <PropertyValue type="object">
+            <Example defaultValue>
+              {`
+{
+    gap: 'medium'
+},
```
jcfilben, line 134: Remove trailing comma here
```suggestion
{
    gap: 'medium'
}
```

PR #302, opened 2021-08-24
Add eslint-plugin-grommet and fix lint errors

taysea:
This PR adds the recommended configuration of `eslint-plugin-grommet` to grommet site. Additionally, I resolved the errors that were flagged by the linter (primarily missing alt-text on Images).

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,6 +1,6 @@
 {
   "parser": "babel-eslint",
-  "extends": ["airbnb", "prettier"],
+  "extends": ["airbnb", "prettier", "plugin:grommet/recommended"],
   "plugins": ["prettier"],
   "env": {
     "node": true,
--- a/src/screens/Home/Message.js
+++ b/src/screens/Home/Message.js
@@ -2,7 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { Box, Heading, Image, Paragraph } from 'grommet';
 
-const Message = ({ imageSrc, name, summary }) => (
+const Message = ({ alt, imageSrc, name, summary }) => (
   <Box basis="medium" align="center">
     <Box
       height="xsmall"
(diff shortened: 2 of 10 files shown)


ericsoderberghp (approved):
Nice to see it doing its thing!

.eslintrc
```diff
@@ -1,6 +1,6 @@
 {
   "parser": "babel-eslint",
-  "extends": ["airbnb", "prettier"],
+  "extends": ["airbnb", "prettier", "plugin:grommet/recommended"],
   "plugins": ["prettier"],
```
halocline, line 4: So this accesses the `recommended` property of the module config from `lib/index.js`?

Is the reason it gets placed in `extends` vs. `plugins` is it is auto applying? Whereas in `plugins` you are opting in a-la-carte?

src/screens/Home/Message.js
```diff
@@ -12,7 +12,7 @@ const Message = ({ imageSrc, name, summary }) => (
       align="center"
       justify="center"
     >
-      <Image width={96} fit="cover" src={imageSrc} a11yTitle={name} />
+      <Image width={96} fit="cover" src={imageSrc} a11yTitle={name} alt={alt} />
```
halocline, line 15: 👏

src/screens/Home/Tools.js
```diff
@@ -31,12 +31,13 @@ const Tile = ({ desc, label, link, src }) => {
         {desc}
       </Paragraph>
       <Box flex />
-      <Image width={imgWidth} src={src} a11yTitle={label} />
+      <Image width={imgWidth} src={src} a11yTitle={label} alt={alt} />
```
halocline, line 34: 👏

halocline (approved):
Nice.

.eslintrc
halocline: Should this be added to the [README.md](https://github.com/grommet/eslint-plugin-grommet/blob/master/README.md)?

taysea: Good call, I'll add some guidance in the README explaining this and how to use.

PR #314, opened 2021-09-22
add lighter and bolder to docs

britt6612:
Add `bolder` and `lighter` to docs for Text weight

### DIFF
--- a/src/screens/Text.js
+++ b/src/screens/Text.js
@@ -197,8 +197,10 @@ export default () => (
         <Property name="weight">
           <Description>Font weight</Description>
           <PropertyValue type="string">
+            <Example>"lighter"</Example>
             <Example>"normal"</Example>
             <Example>"bold"</Example>
+            <Example>"bolder"</Example>
           </PropertyValue>
           <PropertyValue type="number">
             <Example>300</Example>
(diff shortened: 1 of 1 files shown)


src/screens/Text.js
```diff
@@ -197,8 +197,10 @@ export default () => (
         <Property name="weight">
           <Description>Font weight</Description>
           <PropertyValue type="string">
+            <Example>"lighter"</Example>
             <Example>"normal"</Example>
             <Example>"bold"</Example>
+            <Example>"bolder"</Example>
```
halocline, line 203: unless there was a rationale for the order, keeping aligned with [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight#syntax) starting with most often/familiar then into relative to parent. (sorry, being nit-picky here)

```suggestion
            <Example>"normal"</Example>
            <Example>"bold"</Example>
            <Example>"lighter"</Example>
            <Example>"bolder"</Example>
```

halocline (commented):
just a nit on the order of the presentation

src/screens/Text.js
britt6612: I just was putting it in the order that came to mind lol! I can change though

halocline (approved):

PR #316, opened 2021-10-03
Added the dropProps doc, when truncate='tip'

g4rry420:
Signed-off-by: GurkiranSingh <gurkiransinghk@gmail.com>

Added the dropProps doc, when truncate='tip'.
[PR](https://github.com/grommet/grommet/pull/5637)

### DIFF
--- a/src/screens/Text.js
+++ b/src/screens/Text.js
@@ -190,8 +190,7 @@ export default () => (
           <PropertyValue type="string">
             <Description>
               If truncate = tip, the full text content will be placed in a Tip
-              that will appear on hover. The alignment of tip can be done using
-              the `dropProps`.
+              that will appear on hover.
             </Description>
             <Example>"tip"</Example>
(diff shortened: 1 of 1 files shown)


src/screens/Text.js
```diff
@@ -193,7 +190,8 @@ export default () => (
           <PropertyValue type="string">
             <Description>
               If truncate = tip, the full text content will be placed in a Tip
-              that will appear on hover.
+              that will appear on hover. The alignment of tip can be done using
+              the `dropProps`.
```
jcfilben, line 194:
```suggestion
              that will appear on hover.
```
I'm thinking we should leave the description for `truncate` as is and instead add the `dropProps` to the documentation as a separate prop. In the `dropProps` description we should state that it should only be used when `truncate='tip'`.

g4rry420:
Closing, since we don't need this anymore

PR #319, opened 2021-10-12
Adding `children` to Carousel Docs

kevingill55:
This adds a brief explanation for the `children` attribute expected when using the `<Carousel />` component.

### DIFF
--- a/src/screens/Carousel.js
+++ b/src/screens/Carousel.js
@@ -31,21 +31,18 @@ export default () => (
       name="Carousel"
       availableAt={[
         {
-          url:
-            'https://storybook.grommet.io/?selectedKind=Media-Carousel&full=0&stories=1&panelRight=0',
+          url: 'https://storybook.grommet.io/?selectedKind=Media-Carousel&full=0&stories=1&panelRight=0',
           badge:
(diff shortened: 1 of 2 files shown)


src/screens/Carousel.js
```diff
@@ -99,6 +96,21 @@ export default () => (
...
+          <Description>
+            Array of child nodes to use for slides of the Carousel.
+          </Description>
+          <PropertyValue type="array">
+            <Example>
+              {`
+                <Box>Slide One</Box>
+                <Box>Slide Two</Box>
+                <Box>Slide Three</Box>
+              `}
```
jcfilben, line 109:
```suggestion
              {`[
  <Box>Slide One</Box>,
  <Box>Slide Two</Box>,
  <Box>Slide Three</Box>
]`}
```
I wondering if we should wrap with brackets here since if you were passing children as an array through a named prop (`<Carousel children={} />` vs. `<Carousel> {children} </Carousel>`) you would need to have the brackets.

Also we should get rid of the extra tab spacing. If it is left it it will show up on the docs page with the extra tabs. See other props with multiline strings for examples.

PR #320, opened 2021-10-21
add required object to reflect to feat to prop

britt6612:
add the object for `required` prop to reflect new change in grommet
https://github.com/grommet/grommet/pull/5713

### DIFF
--- a/src/screens/FormField.js
+++ b/src/screens/FormField.js
@@ -180,8 +180,8 @@ export default () => (
         <Property name="required">
           <Description>
             Whether the field is required. If `formField.requiredIndicator` =
-            true in your theme, passing {{ indicator: false }} will not display
-            the required indicator for the given field.
+            true in your theme, passing `indicator: false` will not display the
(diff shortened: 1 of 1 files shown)


src/screens/FormField.js
```diff
@@ -181,8 +178,16 @@ export default () => (
         </Property>
 
         <Property name="required">
-          <Description>Whether the field is required.</Description>
-          <GenericBool />
+          <Description>
+            Whether the field is required. If the required indicator should be
```
taysea, line 182:
```suggestion
            Whether the field is required. If `formField.requiredIndicator` =  true in your theme, passing { indicator: false } will not display the required indicator for the given field.
```

src/screens/FormField.js
```diff
@@ -181,8 +178,16 @@ export default () => (
...
+            shown if provided in theme.
```
taysea, line 183:
```suggestion
```

PR #324, opened 2021-11-03
Update Card docs

taysea:
Adjusts card docs based on updates made in https://github.com/grommet/grommet/pull/5762

### DIFF
--- a/src/screens/Card.js
+++ b/src/screens/Card.js
@@ -62,6 +62,13 @@ export default () => (
           </PropertyValue>
         </Property>
 
+        <Property name="card.container.extend">
+          <Description>
+            Any additional style for the card container.
+          </Description>
+          <GenericExtend />
+        </Property>
+
         <Property name="card.footer">
           <Description>Any valid Box prop for the CardFooter.</Description>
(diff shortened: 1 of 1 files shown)


src/screens/Card.js
```diff
@@ -86,13 +93,6 @@ export default () => (
             <Example>"small"</Example>
           </PropertyValue>
         </Property>
-
-        <Property name="card.hover.container.extend">
```
ericsoderberghp, line 90: Isn't this still possible, via
```
kind={{
    hover: theme.card.hover?.container,
    ...theme.card.container,
}}
```

taysea: The extend does not apply.

jcfilben (commented):
Doc changes look fine. It looks like we need to add this theme prop to themes/base.d.ts in Grommet.

jcfilben:
base.d.ts has been updated to include `card.container.extend` in https://github.com/grommet/grommet/pull/6203

It looks like yarn.lock just needs updated and then we can merge this

jcfilben:
Looks like there is still a merge conflict with yarn.lock

PR #330, opened 2021-11-10
DataTable documentation for onUpdate and groupBy changes

MikeKingdom:
This adds the documentation for DataTable's onUpdate addition and groupBy changes
to support lazy loading, especially of group data.

Part of feature [#5679](https://github.com/grommet/grommet/issues/5679)
for PR [#5716](https://github.com/grommet/grommet/pull/5716)

### DIFF
--- a/src/screens/DataTable.js
+++ b/src/screens/DataTable.js
@@ -364,7 +364,7 @@ export default () => (
             <Description disableMarkdown>
               {`This function
             will be called with an array of primary key values, suitable to be
-            passed to the 'select' property and the data item that was changed.
+            passed to the 'select' property, along with the data item that was changed.
             If you are storing select state via
             a 'useState' hook, you can do something like: '
             <DataTable select={select} onSelect={setSelect} />'.`}
@@ -393,7 +393,7 @@ export default () => (
             to the end of existing data. `onUpdate` is also called when the sort
             changes or a group is expanded or collapsed by the user. It is
             passed information on the current sort settings and the maximum
-            number of items it needs to return. If the `groupBy` is specified,
+            number of items it needs to return. If `groupBy` is specified,
             `data` should include an item for each group also. This item should
             have its own unique value in the property specified as the
             `primaryKey` and this unique key value should be listed in
--- a/src/screens/DataTable.js
+++ b/src/screens/DataTable.js
@@ -28,21 +28,18 @@ export default () => (
       name="DataTable"
       availableAt={[
         {
-          url:
-            'https://storybook.grommet.io/?selectedKind=Visualizations-DataTable&full=0&stories=1&panelRight=0',
+          url: 'https://storybook.grommet.io/?selectedKind=Visualizations-DataTable&full=0&stories=1&panelRight=0',
           badge:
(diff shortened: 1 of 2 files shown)


src/screens/DataTable.js
```diff
@@ -275,14 +272,30 @@ export default () => (
               If object is specified 'property' is used to group data by,
               'expand' accepts array of group keys that sets expanded groups and
               'onExpand' is a function that will be called after expand button
-              is clicked with an array of keys of expanded groups.
+              is clicked with an array of keys of expanded groups. If 'onUpdate'
+              is being used to lazy load group items, 'expandable' can be used
+              to specify which items in the data are group header items and
+              'select' and 'onSelect' can be specified in 'groupBy' to help
```
taysea, line 278:
```suggestion
              'select' and 'onSelect' can be used to
```

src/screens/DataTable.js
```diff
@@ -275,14 +272,30 @@ export default () => (
...
+}
+            `}
+            </Example>
+            <Example>
+              {`
+{
+  "property": "location",
+  "expand": ["Paris", "Los Angeles"],
+  "expandable": ["Paris", "Los Angeles", "Fort Collins", "San Jose"],
+  "select": { "": "some", "Paris": "all", "Los Angeles": "some"}
```
taysea, line 297: What does empty string refer to, the entire table? was this chosen to avoid any potential conflicts with user data?

wondering if we can be more clear about this in the description above about how to use `select` in this way

src/screens/DataTable.js
```diff
@@ -351,11 +364,12 @@ export default () => (
             <Description disableMarkdown>
               {`This function
             will be called with an array of primary key values, suitable to be
-            passed to the 'select' property. If you are storing select state via
+            passed to the 'select' property and the data item that was changed.
```
taysea, line 367:
```suggestion
            passed to the 'select' property, along with the data item that was changed.
```

src/screens/DataTable.js
```diff
@@ -370,6 +384,33 @@ export default () => (
...
+        <Property name="onUpdate">
+          <Description disableMarkdown>
+            This function provides a way to lazy-load data. It is similar to
+            `onMore` except it also allows for lazily providing expanded group
+            data and assumes the `data` is already in sorted order. Like
+            `onMore`, DataTable will call `onUpdate` when the user has scrolled
+            to the end of existing data. `onUpdate` is also called when the sort
+            changes or a group is expanded or collapsed by the user. It is
+            passed information on the current sort settings and the maximum
+            number of items it needs to return. If the `groupBy` is specified,
```
taysea, line 396:
```suggestion
            number of items it needs to return. If `groupBy` is specified,
```

src/screens/DataTable.js
MikeKingdom: I put the groupBy part in there to distinguish it from the select/onSelect props themselves. This is a different one and it's a little confusing in text that they have the same name.

src/screens/DataTable.js
MikeKingdom: Yes and yes. This is something we didn't think about and I ran into during implementation. We can probably talk about this more and come up with some better descriptions if nothing else.

taysea (approved):

ericsoderberghp (approved):

PR #334, opened 2021-11-18
Initial Tag component documentation

MikeKingdom:
Closes grommet/hpe-design-system#2059

Tag grommet component PR [#5778](https://github.com/grommet/grommet/pull/5778)


src/screens/Tag.js
```diff
@@ -0,0 +1,230 @@
...
+        <Property name="name">
+          <Description>Name text to place in the tag.</Description>
+          <PropertyValue type="string">
+            <Example>"name"</Example>
+          </PropertyValue>
+        </Property>
+        <Property name="onClick">
+          <Description>
+            Click handler. If this property is specified the tag will be a
+            button.  This should not be used in conjunction with 'onRemove'.
```
taysea, line 90:
```suggestion
            button.  For accessibility, this should not be used in conjunction with 'onRemove' to avoid nested interactive elements.
```
Wondering if we should enhance this to just give a little more education on why you shouldn't use them together.

src/screens/Tag.js
```diff
@@ -0,0 +1,230 @@
...
+          </Description>
+          <PropertyValue type="function">
+            <Example>{`() => {}`}</Example>
+          </PropertyValue>
+        </Property>
+        
+        <Property name="onRemove">
+          <Description>
+            If specified, the tag will include a remove button. This function
+            will be called when the remove button is activated. This should not
```
taysea, line 100:
```suggestion
            will be called when the remove button is activated. For accessibility, this should not be used in conjunction with 'onClick' to avoid nested interactive elements.
```

src/screens/Tag.js
```diff
@@ -0,0 +1,230 @@
...
+            be used in conjunction with 'onClick'.
```
taysea, line 101:
```suggestion
```

src/screens/Tag.js
```diff
@@ -0,0 +1,230 @@
...
+  horizontal: "small",
+  vertical: "xsmall",
+}`}
+            </Example>
+          </PropertyValue>
+        </Property>
+
+        <Property name="tag.remove">
+          <Description>
+            The properties on the remove button
```
taysea, line 187:
```suggestion
            Any valid Box props for the remove button.
```

src/screens/Tag.js
```diff
@@ -0,0 +1,230 @@
...
+            The separator text between the 'name' and 'value' text.
+          </Description>
+          <PropertyValue type="string">
+            <Example defaultValue>":"</Example>
+          </PropertyValue>
+        </Property>
+
+        <Property name="tag.value">
+          <Description>
+            An object specifying the properties for the value Text
```
taysea, line 211:
```suggestion
            Any valid Text props for the value.
```

src/screens/Tag.js
```diff
@@ -0,0 +1,230 @@
...
+              <Example>"top-right"</Example>
+              <Example>"bottom-left"</Example>
+              <Example>"bottom-right"</Example>
+            </PropOptions>
+          </PropertyValue>
+        </Property>
+
+        <Property name="tag.name">
+          <Description>
+            An object specifying the properties for the name Text
```
taysea, line 159:
```suggestion
            Any valid Text props for the name.
```

src/structure.js
```diff
@@ -17,7 +17,7 @@ export const structure = {
     },
     {
       name: 'Type',
-      components: ['Heading', 'Markdown', 'Paragraph', 'Text'],
+      components: ['Heading', 'Markdown', 'Paragraph', 'Text', 'Tag'],
```
taysea, line 20: wondering about the category here -- it feels to me like it aligns more with the "Visualizations"

src/screens/Tag.js
MikeKingdom: We don't tend to explain design decisions. It seems a bit much. Maybe see what others think?

src/screens/Tag.js
MikeKingdom: Why remove it here?

src/screens/Tag.js
MikeKingdom: Aren't these Button props?

src/structure.js
MikeKingdom: It really didn't seem like a Visualization like a chart or graph. It seems more like Text treatment/styling and sometimes a Control. I don't see an obvious answer here yet.

taysea: I think wherever "Tag" goes, then NameValueList should probably follow. Right now we have NameValueList in Visualizations but maybe it should change as well?

src/screens/Tag.js
taysea: My comment above adjusted the whole sentence -- since this was broken onto its own line, it's now repeated with my suggestion above. Only commit this if we accept my above suggestion.

src/screens/Tag.js
taysea: That's fine, I am good leaving it off to stick w convention.

src/screens/Tag.js
taysea: Thanks, adjusted my wording. Pulled this phrasing from other components.

src/structure.js
ericsoderberghp: Might be interesting to run a quick poll via Slack?

britt6612: I think running a poll is a good idea get everyone's thought but I agree where `tags` lives `nameValueList` should also live in the same place

ericsoderberghp (approved):

PR #336, opened 2021-12-01
Expand description for Grommet's full='min' prop

jcfilben:
Provide a more detailed description for the `full='min'` prop. Related to https://github.com/grommet/grommet/pull/5811

### DIFF
--- a/src/screens/Grommet.js
+++ b/src/screens/Grommet.js
@@ -106,13 +106,21 @@ export default () => (
 
         <Property name="full">
           <Description>Whether to take the whole viewport.</Description>
-          <GenericBoolFalse />
+          <PropertyValue type="boolean">
+            <Description>
+              'true' should be used when the entire page layout will be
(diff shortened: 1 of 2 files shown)


src/screens/Grommet.js
```diff
@@ -113,7 +110,10 @@ export default () => (
           <PropertyValue type="string">
             <Description>
               'min' indicates it should take at minimum the whole viewport,
-              allowing the viewport to control scrolling.
+              allowing the viewport to control scrolling. This should be used in
```
ericsoderberghp, line 113: I think this is a little too narrow.

I think `true` should be used when the entire page layout will be controlled by a child Box or Grid, typically with a fixed header, footer, and/or sidebar, with scrolling being handled by the Main content within.

I think `"min"` should be used when the scrolling behavior includes all of the content on the page.

ericsoderberghp (approved):

PR #338, opened 2021-12-12
"screen resolution" -> "viewport size"

matthewoates:
What the page is calling `screen resolution` here is actually the size of the viewport.

Readers might incorrectly think that this is referring to the resolution of the screen (eg: 1920x1080) or that this has something to do with `window.devicePixelRatio`.

Viewport size is the dimensions of the browser tab without the chrome. (address bar, dev tools, etc)

Screen resolution is the resolution of the display. (note: this might not map 1:1 to the number of pixels on the display)

EDIT: I'm not sure if terminology elsewhere needs to be updated - I'm new to Grommet.


jcfilben:
Hi @matthewoates can you try updating you branch with the master branch? Looks like some of the checks are failing on this PR

matthewoates:
@jcfilben updated!

jcfilben (approved):
Thank you, looks good!

PR #349, opened 2022-02-01
add docs for toast position

britt6612:
Add new `position` for Toast

### DIFF
--- a/src/screens/Notification.js
+++ b/src/screens/Notification.js
@@ -133,19 +133,16 @@ return (
         <Property name="toast">
           <Description>
             A boolean flag that indicates whether or not the Notification is a
-            Toast Notification. When using Toast Notification a position is
-            avalible to pass in for where the toast will show.
+            Toast Notification.
           </Description>
           <GenericBoolFalse />
           <PropertyValue type="object">
-            <Example>
-              {`
-{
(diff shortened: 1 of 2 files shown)


src/screens/Notification.js
```diff
@@ -133,11 +133,20 @@ return (
         <Property name="toast">
           <Description>
             A boolean flag that indicates whether or not the Notification is a
-            Toast Notification.
+            Toast Notification. When using Toast Notification a position is
+            avalible to pass in for where the toast will show.
```
jcfilben, line 137: I would put the info about position in the description under the PropertyValue object section instead of the general description for toast since it only applies when passing an object to the toast prop

src/screens/Notification.js
```diff
@@ -133,19 +133,16 @@ return (
...
-{
-  toast: {{ position: 'bottom-right' }},
-}
-              `}
-            </Example>
+            <Example>{`{ toast: {{ position: 'bottom-right' }},}`}</Example>
           </PropertyValue>
+          <Description>
+            When using Toast Notification a position is avalible to pass in as
+            in object which will then be the position of the toast content.
```
jcfilben, line 144:
```suggestion
            an object which will then be the position of the toast content.
```

jcfilben (approved):

PR #354, opened 2022-03-01
Removed beta mode message from DataChart docs

jcfilben:


src/screens/DataChart.js
```diff
@@ -32,10 +32,7 @@ const Examples = () => (
       background="light-3"
     >
       <Paragraph margin="none">
```
ericsoderberghp, line 34: I think we can just remove this whole Paragraph. It feels a bit misleading to include an API surface change in a non-major version without explaining the "beta" concept.

ericsoderberghp (approved):

PR #359, opened 2022-03-24
Update Select docs to show valueLabel can be a function

jcfilben:
Add function as an option for `valueLabel` and fix `valueKey` doc to show both object and function as options.

Associated grommet PR: https://github.com/grommet/grommet/pull/6020

### DIFF
--- a/src/screens/Select.js
+++ b/src/screens/Select.js
@@ -492,6 +492,12 @@ const SelectPage = () => (
           <PropertyValue type="node">
             <Example>{`<Box>...</Box>`}</Example>
           </PropertyValue>
+          <PropertyValue type="function">
+            <Description>
+              A function that returns either a string or a React node.
+            </Description>
+            <Example>{`{(option) => {}}`}</Example>
+          </PropertyValue>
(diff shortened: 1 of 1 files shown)


src/screens/Select.js
```diff
@@ -506,20 +512,24 @@ const SelectPage = () => (
...
-            `}
-            </Example>
+            `}</Example>
+          </PropertyValue>
+          <PropertyValue type="function">
+            <Description>
+              If a function is provided, it is called with the option and should
+              return the value.
+            </Description>
+            <Example>{`(option) => { return "string" }`}</Example>
```
ericsoderberghp, line 532: Is "string" correct here? I'm thinking about cases where the value is a number.

jcfilben: I think returning an object is also valid. I will update the example to just be `(option) => {}` since in the description it states that the function should return the value. Hopefully this is clear enough and the user can look at the documentation for the value prop to see the accepted types.

ericsoderberghp (approved):

PR #360, opened 2022-03-25
Fix/drop-memo-doc: updated doc for memoizing align

amandadupell:
### What this PR does:
- updates guidance around `align` property for both drop and drop button

### DIFF
--- a/src/screens/Drop.js
+++ b/src/screens/Drop.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import { Box } from 'grommet';
+import { Anchor, Box } from 'grommet';
 import Page from '../components/Page';
 import Item from './Components/Item';
 import {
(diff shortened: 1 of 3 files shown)


jcfilben (approved):
Looks good!

src/screens/DropButton.js
```diff
@@ -42,13 +42,20 @@ const DropButtonPage = () => (
...
-      code={`<DropButton
-  label="Fancy Selector"
-  dropAlign={{ top: 'bottom', right: 'right' }}
-  dropContent={
-    <Box pad="large" background="light-2" />
-  }
-/>`}
+      code={`function Example() {
+  const align = { top: "bottom", left: "right" };
+  const dropAlign = React.useMemo(() => align, [align]);
```
ericsoderberghp, line 47: I think this will create a new `align` on each render, and this cause `useMemo` to generate a new `dropAlign` on each render. For this example, I think we can just move `align` outside the render function to a constant at the top of the file.

src/screens/Drop.js
```diff
@@ -49,6 +49,10 @@ const DropPage = () => (
             specifying a vertical or horizontal alignment will cause it to be
             aligned in the center.
           </Description>
+          <Description>
+            The align property object should be memo-ized to maintain the same
```
ericsoderberghp, line 53: This might be a bit too narrow. I think the general concept is to not create a new object on each render but to prevent that by using a persistent object, either as a static constant outside the render path or via `useMemo` if using hooks. It might be worth looking for a React article explaining this and including a link as well.

src/screens/DropButton.js
amandadupell: yes, when testing i realized this `useMemo` still causes the rerender. i've rewritten it so that it is only called when rendered and tested it on the example from [this isses](https://github.com/grommet/grommet/issues/5928) and it is fixing the rerender.

should we recommend this way or moving the prop outside of the render as a constant? i've updated the stories in storybook in this way: https://github.com/grommet/grommet/pull/6017

ericsoderberghp: What about just omitting alignment from the code and relying on the default?

ericsoderberghp (approved):

PR #369, opened 2022-04-06
Fix typo in DataChart granularity

notcarbide:
Discovered while trying example at:
https://v2.grommet.io/datachart


jcfilben (approved):
Nice catch!

PR #370, opened 2022-04-10
Update NameValueList docs

conartist6:
See https://github.com/grommet/grommet/pull/6058


src/screens/NameValueList.js
```diff
@@ -34,7 +34,7 @@ const NameValueListPage = () => (
         },
       ]}
       description="A name-value list displays a group of name-value pairs"
-      intrinsicElement="div"
+      intrinsicElement="dd"
```
ShimiSun, line 37: I think that this change may impact the instinct element type on the List prop type interface.

conartist6: How?

This isn't actually a change to the library, the docs are just wrong at present (as far as I can see). Notice that a `NameValueList` is a `Grid` rendered as a `dl`: https://github.com/grommet/grommet/blob/master/src/js/components/NameValueList/NameValueList.js#L56

conartist6: Now that I write that out I realize that this should have been `dl` not `dd`. Updated the PR.

conartist6: I'm sorry, when you say the "List prop type interface" I guess I'm not sure what you mean. There is `components/NameValueList/propTypes.js`, which perhaps should also be expanded to include `GridPropTypes` in the other PR. Is that what you meant?

jcfilben: I agree with this being updated to `dl`, having the intrinsic element as `div` is incorrect for NameValueList

jcfilben:
Can you try merging your branch with master and running `rm -rf yarn.lock && rm -rf node_modules && yarn cache clean && yarn install`. I think that should resolve the failing checks

conartist6:
Oh. I actually just made this PR with Github's "edit this file" feature.

conartist6:
Built locally. No need to `yarn cache clean`. Why are you telling people to do that?? It's not a good idea...

jcfilben:
> Built locally. No need to yarn cache clean. Why are you telling people to do that?? It's not a good idea...

Since this project uses the stable branch of grommet I've found that sometimes running `yarn cache clean` is necessary to get the latest version of the stable branch. Is there a reason you would not recommend running `yarn cache clean`?

jcfilben (approved):
Looks good!

ericsoderberghp (approved):

conartist6:
Well `cache clean` cleans out your cache. Npm now requires a `--force` flag to clean the cache since they are the ones paying for the bandwidth you'll need to fill it up again. I think that's a bit silly on their part as their messaging suggests there's no reason you should ever need to do it, but in practice a number of real scenarios particularly around using branches as packages can cause cache issues. I'm actually quite interested in that because I'm encouraging people to use git commits and branches as packages more often. As far as I know both `yarn@2` and `pnpm` are able to handle git packages without caching issues, but `yarn@1` and `npm` are not.

conartist6:
I've been shifting towards `pnpm` because `yarn@1` is dead and I don't like how `yarn@2` wants the history of their package manager to be part of the history of every single git repo.

jcfilben:
Thanks for the info!
Looks like there is a conflict with yarn.lock on this PR, can you merge with master and regenerate the yarn.lock file?

conartist6:
I've rebased a second time. You know I'd think it would be easy to make sure people don't have to do this wouldn't it? `yarn` itself is capable of resolving conflicts in its lockfile, I think you just need to configure it as a merge driver: https://github.com/Praqma/git-merge-driver

PR #375, opened 2022-04-27
Add doc changes for PR's before grommet release

jcfilben:
Added necessary doc updates before releasing next version of grommet. Changes are related to the following pull requests:
https://github.com/grommet/grommet/pull/6038 - Make `as` prop on Box accept any React element
https://github.com/grommet/grommet/pull/6076 - Allow DataTable vertical align to be specified by prop

### DIFF
--- a/src/screens/DataTable.js
+++ b/src/screens/DataTable.js
@@ -632,6 +632,26 @@ const DataTablePage = () => (
             <Example defaultValue>50</Example>
           </PropertyValue>
         </Property>
+
+        <Property name="verticalAlign">
+          <Description>How to vertically align items.</Description>
+          <PropertyValue type="string">
+            <Example>"bottom"</Example>
+            <Example>"middle"</Example>
+            <Example>"top"</Example>
+          </PropertyValue>
+          <PropertyValue type="object">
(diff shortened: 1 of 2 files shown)


src/screens/DataTable.js
```diff
@@ -192,7 +192,8 @@ const DataTablePage = () => (
               first column will be used. 'pin' indicates that this column should
               not scroll out of view to the left when the table is scrolled
               horizontally. 'plain' = true indicates that the body cells in the
-              column will not apply pad.
+              column will not apply pad. Setting `verticalAlign` within columns
```
ericsoderberghp, line 195: Just checking whether it is "deprecated" vs. whether it is now available at a table-wide level in addition.

jcfilben: Hmm I looked back at Taylor's PR and it looks like she added a comment that it should be deprecated in v3... So not deprecated yet. I will go back and remove the deprecated tuff from the doc.

jcfilben: Fixed in latest commit

ericsoderberghp (approved):

PR #392, opened 2022-06-24
Add doc changes for grouped menu

jcfilben:
Added doc updates for https://github.com/grommet/grommet/pull/6162 and https://github.com/grommet/grommet/pull/6201

### DIFF
--- a/src/screens/Menu.js
+++ b/src/screens/Menu.js
@@ -209,7 +209,7 @@ const MenuPage = () => (
           <PropertyValue type="array[array[object]]">
             <Description>
               Items can be defined as an array of arrays of objects to create
-              groups of menu items.
+              groups of Menu items.
             </Description>
             <Example> {`[[{}], [{}]]`} </Example>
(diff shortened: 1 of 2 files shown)


src/screens/Menu.js
```diff
@@ -206,6 +206,13 @@ const MenuPage = () => (
           <PropertyValue type="array[object]">
             <Example> {`[{}, {}]`} </Example>
           </PropertyValue>
+          <PropertyValue type="array[array[object]]">
+            <Description>
+              Items can be defined as an array of arrays of objects to create
+              groups of Menu items.
+            </Description>
+            <Example> {`[[{}], [{}]]`} </Example>
```
taysea, line 214: Would it be more valuable to include more detail about what that object might look like? (And as a bonus, improve the non-group menu example as well)

jcfilben: Good idea, I added this in the latest commit

taysea (approved):

Issue #397, opened 2022-07-06
Add link to HPE Design System docs to component pages


taysea:
builds off of #402 

Right now, there is low awareness about how to transition from Grommet docs to HPE Design System guidance on a component. In order to raise visibility of HPE Design System docs and examples, a link should be added to each component page (alongside the Storybook, CodeSandbox, Github links) that leads to the Design System guidance for that component.

Deliverables:
- [ ] Adjust site layout to allow for "Design System Guidance" link when on HPE theme (tbd on label)
- [ ] Add links to each component page

ericsoderberghp:
I think we should do this only when the theme is `hpe`. And, we might want to consider making it easier to switch the theme of the grommet documents as well.

JCoder121:
After meeting with Jessica, I will first work on different prototypes for an easier way to toggle between themes. We will evaluate these prototypes and then move forward with adding links to HPE Design System docs to components page.

JCoder121:
Currently most updated direction: see #405.

PR #398, opened 2022-07-09
Add HPE/Grommet theme switch button (icon) prototype

JCoder121:
Closed in favor of #399. 

image
image

This prototype uses the HPE/Grommet icons as buttons for the user to switch between the HPE theme and the Grommet theme across the Grommet site. Note: Jessica brought up potential legal issues with including HPE logo on open source website, need to confirm. 
Related to Issue #397.

### DIFF
--- a/src/components/Nav.js
+++ b/src/components/Nav.js
@@ -1,6 +1,6 @@
 import React from 'react';
-import { Box, Text, ResponsiveContext } from 'grommet';
-import { Grommet as GrommetIcon } from 'grommet-icons';
+import { Box, Text, ResponsiveContext, Button } from 'grommet';
+import { Grommet as GrommetIcon, Hpe as HpeIcon } from 'grommet-icons';
 import RoutedAnchor from './RoutedAnchor';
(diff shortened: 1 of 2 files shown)


src/components/Nav.js
```diff
@@ -51,6 +51,34 @@ const Nav = () => {
           </Box>
         )}
         <Search open={searchOpen} setOpen={(value) => setSearchOpen(value)} />
+        <Button
+          tip={
+            window.location.href.indexOf('hpe') > -1
```
MikeKingdom, line 56: Seems like this should come from the theme context or at least the `themeName` stored in App.js. Since it's stateful in the App that means it may not actually always be in the `window.location`. It's also possible for it to be something other than `hpe`, `grommet` or undefined (like aruba, dark, dhc, hp...). 

And picking out a query param should instead do this by actually looking for the `theme` query parameter. App.js shows how to pick out query parameters.

I'm surprised we haven't tweaked this to use the react router. Then we could use the `useParams()` hook :)

JCoder121:
Closed in favor of #399

PR #399, opened 2022-07-13
Add HPE/Grommet theme switch button (text) prototype

JCoder121:
Waiting on further design guidance before moving forward. [Not moving forward with theme switch button, may re-open this PR in the future.]

[Deploy Preview](https://deploy-preview-399--upbeat-ardinghelli-95952e.netlify.app/)

image

image

image

image

image

This prototype uses a button in the Nav section of each page on the Grommet site as a theme switcher. Updated code to use Context rather than depend on the URL to supply/hardcode initial theme values, so each button click doesn't change the search URL and instead re-skins the content according to the new theme. Button is dynamically changed into a menu and text next to Grommet icon is hidden to reduce footprint in Nav bar when resizing to `small` and `xsmall` window sizes. Final visual design for Theme Switch Button and theme description under Grommet icon TBD. Note: URL with ?theme behavior still need to be discussed.
Related to Issue #397.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -24,12 +24,12 @@ const THEMES = {
 
 const App = ({ initialPath }) => {
   const [themeName, setThemeName] = React.useState(() => {
-    // Get stored value from local storage, 'grommet' as default
+    // Get stored value from local storage, set to 'grommet' as default
+    let storedTheme;
     if (typeof window !== 'undefined') {
-      const storedTheme = JSON.parse(window.localStorage.getItem('theme-name'));
-      return storedTheme;
+      storedTheme = JSON.parse(window.localStorage.getItem('theme-name'));
     }
-    return 'grommet';
+    return storedTheme || 'grommet';
   });
   const [search, setSearch] = React.useState();
 
@@ -72,7 +72,7 @@ const App = ({ initialPath }) => {
       </Helmet>
       <Analytics>
         <ThemeSwitchContext.Provider value={themeSwitchContextValue}>
-          <Grommet theme={THEMES[themeName || 'grommet']}>
+          <Grommet theme={THEMES[themeName]}>
             <Content />
           </Grommet>
         </ThemeSwitchContext.Provider>
--- a/src/App.js
+++ b/src/App.js
@@ -11,6 +11,7 @@ import { dxc } from 'grommet-theme-dxc';
 import { Router } from './Router';
 import Analytics from './components/Analytics';
 import Content from './components/Content';
+import { ThemeSwitchContext } from './components/ThemeSwitchContext';
 
 const THEMES = {
   grommet,
@@ -36,6 +37,11 @@ const App = ({ initialPath }) => {
     }
   }, []);
 
+  const themeSwitchContextValue = React.useMemo(() => {
+    const nextValue = { themeName, setThemeName };
+    return nextValue;
+  }, [themeName]);
+
   return (
     <Router initialPath={initialPath} search={search}>
       <Helmet titleTemplate="%s - Grommet" defaultTitle="Grommet">
@@ -46,9 +52,11 @@ const App = ({ initialPath }) => {
         />
       </Helmet>
       <Analytics>
-        <Grommet theme={THEMES[themeName || 'grommet']}>
-          <Content />
-        </Grommet>
+        <ThemeSwitchContext.Provider value={themeSwitchContextValue}>
+          <Grommet theme={THEMES[themeName || 'grommet']}>
+            <Content />
+          </Grommet>
+        </ThemeSwitchContext.Provider>
       </Analytics>
     </Router>
   );
--- a/src/App.js
+++ b/src/App.js
@@ -23,19 +23,35 @@ const THEMES = {
 };
 
 const App = ({ initialPath }) => {
-  const [themeName, setThemeName] = React.useState('grommet');
+  const [themeName, setThemeName] = React.useState(() => {
+    // get stored value from local storage, grommet as default
+    const storedTheme = JSON.parse(window.localStorage.getItem('theme-name'));
+    return storedTheme || 'grommet';
+  });
   const [search, setSearch] = React.useState();
 
   React.useEffect(() => {
+    // note: json.stringify is optional when working with actual string (like theme name)
     if (window.location.search) {
       const {
         location: { search: nextSearch },
       } = window;
       const params = new URLSearchParams(nextSearch);
       setSearch(nextSearch);
-      setThemeName(params.get('theme'));
+      const paramsTheme = params.get('theme');
+      // QUESTION: CURRENT IMPLEMENTATION = re-skin, but with theme in params, when reload, what happens?
+      if (paramsTheme) {
+        setThemeName(paramsTheme);
+      } else {
+        setThemeName(themeName);
+      }
     }
   }, []);
+
+  // use effect hook AFTER setting theme, if can retrieve. optimize performance?
+  React.useEffect(() => {
+    window.localStorage.setItem('theme-name', JSON.stringify(themeName));
+  }, [themeName]);
 
   const themeSwitchContextValue = React.useMemo(() => {
     const nextValue = { themeName, setThemeName };
--- a/src/App.js
+++ b/src/App.js
@@ -24,14 +24,13 @@ const THEMES = {
 
 const App = ({ initialPath }) => {
   const [themeName, setThemeName] = React.useState(() => {
-    // get stored value from local storage, grommet as default
+    // Get stored value from local storage, 'grommet' as default
     const storedTheme = JSON.parse(window.localStorage.getItem('theme-name'));
     return storedTheme || 'grommet';
   });
   const [search, setSearch] = React.useState();
 
   React.useEffect(() => {
-    // note: json.stringify is optional when working with actual string (like theme name)
     if (window.location.search) {
       const {
         location: { search: nextSearch },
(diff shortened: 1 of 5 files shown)


src/App.js
```diff
@@ -32,10 +40,27 @@ const App = ({ initialPath }) => {
...
+        setThemeName(paramsTheme);
+      } else {
+        setThemeName(themeName);
+      }
     }
   }, []);
 
+  React.useEffect(() => {
+    if (typeof window !== 'undefined') {
+      window.localStorage.setItem('theme-name', JSON.stringify(themeName));
```
taysea, line 55: On first render, the cookie isn't getting assigned so I'm seeing something like this (missing descriptive text and missing icon from hpe button)

Screen Shot 2022-07-15 at 2 51 22 PM

src/components/Nav.js
```diff
@@ -18,14 +20,54 @@ const Nav = () => {
       alignSelf="center"
       gap="medium"
     >
-      <RoutedAnchor
-        path="/"
-        icon={<GrommetIcon size="large" />}
-        label={size !== 'small' && <Text size="xlarge">grommet</Text>}
-      />
+      <Box
+        direction="column"
```
taysea, line 24:
```suggestion
```
Column is default direction, so we can remove I think

src/components/Nav.js
```diff
@@ -18,14 +20,54 @@ const Nav = () => {
...
+        justify="between"
+        align="start"
+        alignSelf="center"
+        width="medium"
```
taysea, line 28:
```suggestion
```
Do we need to set a restricted width? or can we just let the content define how big the Box is.

src/components/Nav.js
```diff
@@ -18,14 +20,54 @@ const Nav = () => {
...
-      <RoutedAnchor
-        path="/"
-        icon={<GrommetIcon size="large" />}
-        label={size !== 'small' && <Text size="xlarge">grommet</Text>}
-      />
+      <Box
+        direction="column"
+        justify="between"
+        align="start"
+        alignSelf="center"
```
taysea, line 27: Do we need the justify, align, and alignSelf props? or can we remove.

src/components/Nav.js
```diff
@@ -18,14 +20,54 @@ const Nav = () => {
...
+        <RoutedAnchor
+          path="/"
+          icon={<GrommetIcon size="large" />}
+          label={size !== 'small' && <Text size="xlarge">grommet</Text>}
+        />
+        {themeName === 'grommet' && <Text size="small">for open source</Text>}
+        {themeName === 'hpe' && <Text size="small">for HPE</Text>}
+      </Box>
+
       <Box direction="row" gap="small">
```
taysea, line 39: It seems like this row of buttons should be aligned to the top of the row -- can you look at adjusting the outer Box of the nav to have align=start?

src/components/Nav.js
```diff
@@ -18,14 +20,54 @@ const Nav = () => {
...
+                <Box
+                  pad={{ vertical: 'small', horizontal: 'small' }}
+                  round="xlarge"
+                  background={hover ? 'active' : undefined}
+                  direction="row"
+                  gap="xsmall"
+                >
+                  <Text>
+                    {themeName === 'hpe'
+                      ? 'use for open source'
```
taysea, line 62: Proposing we simplify the logic to be:

`Use for ${themeName === 'hpe' ? HPE : open source}`

src/components/Nav.js
```diff
@@ -18,14 +20,54 @@ const Nav = () => {
...
+        />
+        {themeName === 'grommet' && <Text size="small">for open source</Text>}
+        {themeName === 'hpe' && <Text size="small">for HPE</Text>}
+      </Box>
+
       <Box direction="row" gap="small">
         {!searchOpen && (
           <Box direction="row" gap="small">
+            <Button
+              plain
```
taysea, line 43: When button has children, I think it automatically behaves as `plain`, so we could remove this.

src/components/Nav.js
```diff
@@ -18,14 +20,54 @@ const Nav = () => {
...
+                  direction="row"
+                  gap="xsmall"
+                >
+                  <Text>
+                    {themeName === 'hpe'
+                      ? 'use for open source'
+                      : 'use for HPE'}
+                  </Text>
+                  {themeName === 'grommet' && (
+                    <HpeIcon color="#01A982" size="medium" />
```
taysea, line 66:
```suggestion
                    <HpeIcon color="plain" size="medium" />
```
This icon happens to have the color defined on the SVG, so if we apply `plain` as its color it will just use that.

src/App.js
```diff
@@ -46,9 +71,11 @@ const App = ({ initialPath }) => {
         />
       </Helmet>
       <Analytics>
-        <Grommet theme={THEMES[themeName || 'grommet']}>
-          <Content />
-        </Grommet>
+        <ThemeSwitchContext.Provider value={themeSwitchContextValue}>
+          <Grommet theme={THEMES[themeName || 'grommet']}>
```
taysea, line 75:
```suggestion
          <Grommet theme={THEMES[themeName]}>
```
If themeName always defaults to grommet, we probably don't need this fallback anymore.

src/components/Nav.js
JCoder121: Removed!

src/components/Nav.js
JCoder121: image
When I removed `plain`, this happened. Not sure if there is another fix around it, but would keeping `plain` be the most effective solution?

src/components/Nav.js
JCoder121: image
This is my result when I use plain, might there be another label?

src/components/Nav.js
taysea: oh interesting, okay then leave `plain` !

src/components/Nav.js
taysea: Hmm that's odd, plain should work. You can leave as hex code for now.

JCoder121: I think before I tried using `brand` but that colored in in the Grommet purple, so I ended up just using the hex values.

src/components/Nav.js
```diff
@@ -1,12 +1,19 @@
...
+  Hpe as HpeIcon,
+  More as MoreIcon,
+} from 'grommet-icons';
 import RoutedAnchor from './RoutedAnchor';
 import RoutedButton from './RoutedButton';
 import Search from './Search';
+import { ThemeSwitchContext } from './ThemeSwitchContext';
 
 const Nav = () => {
+  // themename vs theme - make sure to pick
```
taysea, line 15: can we remove this comment?

src/components/Nav.js
```diff
@@ -15,19 +22,70 @@ const Nav = () => {
...
-        icon={<GrommetIcon size="large" color="plain" />}
-        label={size !== 'small' && <Text size="xlarge">grommet</Text>}
-      />
+      <Box justify="between" align="start" alignSelf="center">
+        <RoutedAnchor
+          path="/"
+          icon={<GrommetIcon size="large" color="plain" />}
+          label={
+            !['xsmall', 'small'].includes(size) && (
+              <Text size="xlarge">grommet</Text>
```
taysea, line 36: It feels like this should always be the grommet theme font and color because it's more of a branded element -- can we explore wrapping this Text element in `ThemeContext.Extend` where the value would include the font from grommet theme?
```
global: {
    font: {
      family: `-apple-system,
         BlinkMacSystemFont, 
         "Segoe UI", 
         Roboto, 
         Oxygen, 
         Ubuntu, 
         Cantarell, 
         "Fira Sans", 
         "Droid Sans",  
         "Helvetica Neue", 
         Arial, sans-serif,  
         "Apple Color Emoji", 
         "Segoe UI Emoji", 
         "Segoe UI Symbol"`,
    },
  },
```

https://user-images.githubusercontent.com/12522275/179637751-d1ced39b-16d6-4018-bd72-5c17925e9a21.mov

src/components/Nav.js
```diff
@@ -15,19 +22,70 @@ const Nav = () => {
...
+                {({ hover }) => (
+                  <Box
+                    pad={{ vertical: 'small', horizontal: 'small' }}
+                    round="xlarge"
+                    background={hover ? 'active' : undefined}
+                    direction="row"
+                    gap="xsmall"
+                  >
+                    <Text>
+                      use for {themeName === 'hpe' ? 'open source' : 'HPE'}
```
taysea, line 63:
```suggestion
                      Use for {themeName === 'hpe' ? 'open source' : 'HPE'}
```

src/components/Nav.js
```diff
@@ -15,19 +22,70 @@ const Nav = () => {
...
+                    </Text>
+                    {themeName === 'grommet' && (
+                      <HpeIcon color="#01A982" size="medium" />
```
taysea, line 66: Is medium the default size?

src/components/Nav.js
```diff
@@ -15,19 +22,70 @@ const Nav = () => {
...
+                    )}
+                  </Box>
+                )}
+              </Button>
+            )}
+            {['xsmall', 'small'].includes(size) && (
+              <Menu
+                icon={<MoreIcon size="medium" />}
```
taysea, line 74: is medium default size?

src/components/Nav.js
```diff
@@ -15,19 +22,70 @@ const Nav = () => {
...
+                items={[
+                  {
+                    label: `use for ${
```
taysea, line 77:
```suggestion
                    label: `Use for ${
```

taysea: We are repeating this string logic in the button and the menu, can we extract it to a single `const buttonLabel`?

src/components/Nav.js
JCoder121: Yeah sorry forgot to take that out

src/components/Nav.js
JCoder121: Removed!

jcfilben:
Closing, we ended up going a different direction. See https://github.com/grommet/grommet-site/pull/405

PR #401, opened 2022-07-13
Add HPE/Grommet theme switch menu prototype

JCoder121:
Closed in favor of #399. 

[Deploy Preview](https://deploy-preview-401--upbeat-ardinghelli-95952e.netlify.app/)

image

image

image

image

This prototype uses a menu in the Nav section of each page on the Grommet site as a theme switcher. Context code and logic remain similar to the other prototypes. Upon visiting the site for the first time, the default theme shown in the menu is 'grommet theme'. Final visual design for Theme Switch Menu TBD.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -11,6 +11,7 @@ import { dxc } from 'grommet-theme-dxc';
 import { Router } from './Router';
 import Analytics from './components/Analytics';
 import Content from './components/Content';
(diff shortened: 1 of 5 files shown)


ericsoderberghp:
Based on the screenshots shown, it looks like there will be no theme selected on initial entry and the user can then choose a particular theme if desired. However, there always needs to be an active theme. The site will never have "no theme".

PR #405, opened 2022-07-20
Add HPE documentation links to Grommet component screens

### LINKED ISSUE #397: Add link to HPE Design System docs to component pages

JCoder121:
[Deploy Preview](https://deploy-preview-405--upbeat-ardinghelli-95952e.netlify.app/box)

image

image

Add Notification component allowing user to click a link redirecting to the HPE Design Site's documentation on the corresponding component for further reference. Not all Grommet components have dedicated documentation on DS (ex: Avatar, Sidebar, Color), so the notification will not appear. 
Closes #397.

### DIFF
--- a/src/components/Doc/ComponentDoc.js
+++ b/src/components/Doc/ComponentDoc.js
@@ -110,12 +110,8 @@ export const ComponentDoc = ({
             status="info"
             actions={[
               {
-                label: `See ${name} for examples and guidance.`,
+                label: `See ${name} examples and guidance.`,
                 href: designSystemLink,
-                color: 'text-strong',
-                style: {
-                  textDecorationLine: 'underline',
-                },
               },
             ]}
           />
--- a/src/components/Doc/ComponentDoc.js
+++ b/src/components/Doc/ComponentDoc.js
@@ -90,7 +90,13 @@ export const ComponentDoc = ({
         <Button
           href={`https://design-system.hpe.design/components/${name.toLowerCase()}`}
         >
-          <Text color="black" weight="bold">
+          <Text
+            style={{
+              textDecorationLine: 'underline',
+            }}
+            color="black"
+            weight="bold"
+          >
             See {name} for examples and guidance.
           </Text>
         </Button>
--- a/src/components/Doc/ComponentDoc.js
+++ b/src/components/Doc/ComponentDoc.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import Helmet from 'react-helmet';
 import PropTypes from 'prop-types';
-import { Anchor, Box, Paragraph, Text } from 'grommet';
+import { Anchor, Box, Button, Paragraph, Text } from 'grommet';
 import { Github } from 'grommet-icons';
 import Header from '../Header';
 import RoutedAnchor from '../RoutedAnchor';
@@ -77,6 +77,23 @@ export const ComponentDoc = ({
             )}
           </Box>
         )}
+      </Box>
+
+      <Box
+        direction="row"
+        gap="small"
+        background="light-1"
+        pad="small"
+        round="small"
+      >
+        <Text>Using Grommet for HPE?</Text>
+        <Button
+          href={`https://design-system.hpe.design/components/${name.toLowerCase()}`}
+        >
+          <Text color="black" weight="bold">
+            See {name} for examples and guidance.
+          </Text>
+        </Button>
       </Box>
 
       {isA && (
--- a/src/components/Doc/ComponentDoc.js
+++ b/src/components/Doc/ComponentDoc.js
@@ -112,6 +112,8 @@ export const ComponentDoc = ({
               {
                 label: `See ${name} examples and guidance.`,
                 href: designSystemLink,
+                target: '_blank',
+                rel: 'noreferrer',
               },
             ]}
           />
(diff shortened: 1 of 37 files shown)


src/components/Doc/ComponentDoc.js
```diff
@@ -79,6 +87,41 @@ export const ComponentDoc = ({
...
+              },
+            },
+          }}
+        >
+          <Notification
+            message="Using Grommet for HPE?"
+            status="info"
+            actions={[
+              {
+                label: `See ${name} for examples and guidance.`,
```
taysea, line 113:
```suggestion
                label: `See ${name} examples and guidance.`,
```

src/components/Doc/ComponentDoc.js
```diff
@@ -79,6 +87,41 @@ export const ComponentDoc = ({
...
+                href: designSystemLink,
+                color: 'text-strong',
+                style: {
```
taysea, line 116: @ericsoderberghp I know you aim to avoid inline styling as much as possible, but this feels like the cleanest to getting an underline under the text. Are you okay with this?

src/screens/Calendar.js
```diff
@@ -43,6 +43,7 @@ const CalendarPage = () => (
           label: 'Github',
         },
       ]}
+      designSystemLink="https://design-system.hpe.design/components/calendar"
```
taysea, line 46: let's remove since we don't have a calendar page

src/components/Doc/ComponentDoc.js
ericsoderberghp: Something just feels existentially wrong to me about having inline styles in the Design System site. Can we just start with the default and see how it looks?

taysea: Yes, we can do that. @JCoder121 -- let's remove the inline style of underline and custom text color and just leave as the default grommet anchor treatment.

src/components/Doc/ComponentDoc.js
```diff
@@ -79,6 +87,37 @@ export const ComponentDoc = ({
...
+            },
+          }}
+        >
+          <Notification
+            message="Using Grommet for HPE?"
+            status="info"
+            actions={[
+              {
+                label: `See ${name} examples and guidance.`,
+                href: designSystemLink,
```
taysea, line 114: can we add `target: _blank` and `rel: noreferrer` so that it'll open in a new window?

src/components/Doc/ComponentDoc.js
```diff
@@ -79,6 +87,39 @@ export const ComponentDoc = ({
...
+                target: '_blank',
+                rel: 'noreferrer',
```
taysea, line 116:
```suggestion
```
sorry -- I changed my mind, I think noreferrer could actually be helpful from DS analytics to know how many people are coming from grommet site.

taysea (approved):
LGTM!

src/components/Doc/ComponentDoc.js
```diff
@@ -79,6 +87,38 @@ export const ComponentDoc = ({
         )}
       </Box>
 
+      {designSystemLink && (
+        <ThemeContext.Extend
```
ericsoderberghp, line 91: Why do we need to override the theme? Can we use the default?

JCoder121: I think it was because I couldn't find a way to make the default notification visually match what we agreed on, and it was easier to override the theme and manually specify certain parameters.

taysea: I think the main thing we wanted was to leverage Notification component for layout and text styling, but needed to add background color/rounding/direction that differs from grommet's theme.

https://storybook.grommet.io/?path=/story/visualizations-notification-status--status

taysea: Alternately, Jeffrey had explored just building the entire thing with Box, but getting to leverage Notification felt nice if possible (but does require theme overrides since these things aren't adjustable via props) @ericsoderberghp

taysea: Maybe the default grommet theming looks okay though and we just stick with that?

ericsoderberghp: Semantically, this feels like it should be a Notification, something to call attention to. If the grommet theme isn't working, then perhaps it means we should change the grommet theme. I'm inclined to use the components we've got as much as possible and only tweak them when we need to use them in distinct ways. All that said, I'll leave it up to the two of you to decide how you want to proceed. I'm OK taking this as is and submitting a new issue to re-evaluate the grommet theme for Notification, if preferred.

taysea: I chatted with Aditi and she didn't have any allergic reactions to the grommet default styling. In the spirit of keeping things lean, let's go with the default grommet styling for this first pass.

Next steps after merging -->

We can touch base with some of the developers I interviewed and open-ended let them know we made some small changes to the docs pages and let them browse. If they notice the new link, awesome! If no one notices it, that will indicate visibility is too low and we might need to rethink the visual treatment.

taysea (approved):
LGTM!

ericsoderberghp (approved):

PR #411, opened 2022-08-23
Documents Heading's weight and overflowWrap props

halocline:
Documents Heading's weight and overflowWrap props. Pending merge of https://github.com/grommet/grommet/pull/6296.


jcfilben (approved):
Looks good!

src/screens/Heading.js
```diff
@@ -123,6 +123,17 @@ const HeadingPage = () => (
...
         </Property>
 
+        <Property name="overflowWrap">
+          <Description>
+            Specifies whether line breaks should be inserted within an otherwise
+            unbreakable string to prevent text from overflowing.
+          </Description>
+          <PropertyValue type="string">
+            <Example>"break-word"</Example>
+            <Example>any CSS for overflow-wrap</Example>
```
ericsoderberghp, line 133: Can we show the typical uses inline, rather than requiring the caller to figure out how to look up "any CSS for overflow-wrap"?

ericsoderberghp: I'm thinking both concrete example values plus a sentence for each in the Description.

halocline: Added each value as option and attempted to define each in a brief but understandable manner. Not sure if I succeeded... 😬

ericsoderberghp (approved):

Issue #362, opened 2022-03-28
Show if a component or prop is only available on the stable branch


jcfilben:
There have been issues lately with people trying to use components and props that are only available on the stable branch and not officially release yet. From the documentation there is no way to tell if something is only available on the stable branch. We should have some indication of this.

We will also need to add a step to the release process to go through and remove the only available on stable branch communication from the docs once a component or prop is released.

PR #415, opened 2022-09-08
add directions for how to use stable branch

britt6612:
Add stable docs to site.

PR #416, opened 2022-09-09
Stable notification

jcfilben:
Adds ability to mark components and specific prop as only being available on the stable branch.

https://github.com/grommet/grommet-site/pull/415 should be merged first so the links to the stable branch in this PR go to the stable branch document on the site.

Closes https://github.com/grommet/grommet-site/issues/362

### DIFF
--- a/src/components/Doc/ComponentDoc.js
+++ b/src/components/Doc/ComponentDoc.js
@@ -19,6 +19,7 @@ export const ComponentDoc = ({
   name,
   pageTitle,
   designSystemLink,
+  stable,
 }) => {
   const properties = [];
   const themeDoc = [];
@@ -49,6 +50,22 @@ export const ComponentDoc = ({
         <title>{pageTitle || name}</title>
         <meta name="description" content={description} />
       </Helmet>
+      {stable && (
+        <Box pad={{ bottom: 'medium' }}>
+          <Box
+            alignContent="center"
+            direction="row"
+            background="light-3"
+            pad="xsmall"
+            round="xsmall"
+          >
+            <Text alignSelf="center">
+              <b>New!</b> This component is available on the{' '}
+              <Anchor href="/stable">stable branch</Anchor>
+            </Text>
+          </Box>
+        </Box>
+      )}
       <Box align={align}>
         <Header align={align} label={name} summary={description} details="" />
 
@@ -162,6 +179,7 @@ ComponentDoc.propTypes = {
   name: PropTypes.string.isRequired,
   pageTitle: PropTypes.string,
   designSystemLink: PropTypes.string,
+  stable: PropTypes.bool,
 };
 
 ComponentDoc.defaultProps = {
@@ -174,4 +192,5 @@ ComponentDoc.defaultProps = {
   isA: undefined,
   pageTitle: undefined,
   designSystemLink: undefined,
+  stable: false,
 };
--- a/src/components/Doc/ComponentDoc.js
+++ b/src/components/Doc/ComponentDoc.js
@@ -51,19 +51,21 @@ export const ComponentDoc = ({
         <meta name="description" content={description} />
       </Helmet>
       {stable && (
-        <Notification
-          background="light-3"
-          status="info"
-          global
-          actions={[
-            {
-              label: 'stable branch.',
-              href: '/stablebranch',
-              target: '_blank',
-            },
-          ]}
-          message={`New! ${name} is in beta, test it out and let us know what you think. This component is available on the`}
-        />
+        <Box pad={{ bottom: 'medium' }}>
+          <Notification
+            background="light-3"
+            status="info"
+            global
+            actions={[
+              {
+                label: 'stable branch.',
+                href: '/stablebranch',
+                target: '_blank',
+              },
+            ]}
+            message={`New! ${name} is in beta, test it out and let us know what you think. This component is available on the`}
+          />
+        </Box>
       )}
       <Box align={align}>
         <Header align={align} label={name} summary={description} details="" />
--- a/src/components/Doc/ComponentDoc.js
+++ b/src/components/Doc/ComponentDoc.js
@@ -51,20 +51,19 @@ export const ComponentDoc = ({
         <meta name="description" content={description} />
       </Helmet>
       {stable && (
-        <Box pad={{ bottom: 'medium' }}>
-          <Box
-            alignContent="center"
-            direction="row"
-            background="light-3"
-            pad="xsmall"
-            round="xsmall"
-          >
-            <Text alignSelf="center">
-              <b>New!</b> This component is available on the{' '}
-              <Anchor href="/stable">stable branch</Anchor>
-            </Text>
-          </Box>
-        </Box>
+        <Notification
+          background="light-3"
+          status="info"
+          global
+          actions={[
+            {
+              label: 'stable branch.',
+              href: '/stablebranch',
+              target: '_blank',
+            },
+          ]}
+          message={`New! ${name} is in beta, test it out and let us know what you think. This component is available on the`}
+        />
       )}
       <Box align={align}>
         <Header align={align} label={name} summary={description} details="" />
--- a/src/components/Doc/ComponentDoc.js
+++ b/src/components/Doc/ComponentDoc.js
@@ -55,7 +55,6 @@ export const ComponentDoc = ({
           <Notification
             background="light-3"
             status="info"
-            global
             actions={[
               {
                 label: 'stable branch.',
--- a/src/components/Doc/Property.js
+++ b/src/components/Doc/Property.js
@@ -27,7 +27,7 @@ export const Property = ({ stable, name, first, children }) => {
       onBlur={() => setOver(false)}
     >
       <Box direction="row" justify="between" align="center">
-        <Box direction="row" gap="small">
+        <Box direction="row-responsive" gap="small">
           <Heading level={3} margin="none">
             {name}
           </Heading>
@@ -41,10 +41,11 @@ export const Property = ({ stable, name, first, children }) => {
             >
               <CircleInformation />
               <Text alignSelf="center" size="small">
-                New! Available on{' '}
+                New! Available on the{' '}
                 <Anchor href="/stablebranch" target="_blank">
                   stable branch
                 </Anchor>
+                .
               </Text>
             </Box>
           )}
--- a/src/components/Doc/Property.js
+++ b/src/components/Doc/Property.js
@@ -1,10 +1,10 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { Box, Anchor, Heading } from 'grommet';
-import { Link as LinkIcon } from 'grommet-icons';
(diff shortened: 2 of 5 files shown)


MikeKingdom:
Would it be possible to pass a version number instead?  It would be interesting if you could instead specify what version we expect the ability to be in then the site could do both a 'new in stable' or a 'new in version X' (and only show that if the current version of  grommet is that version.) It might help avoid having to go find and clean up old ones with each grommet release.

src/components/Doc/ComponentDoc.js
```diff
@@ -49,6 +50,22 @@ export const ComponentDoc = ({
         <title>{pageTitle || name}</title>
         <meta name="description" content={description} />
       </Helmet>
+      {stable && (
+        <Box pad={{ bottom: 'medium' }}>
```
halocline, line 54: Potentially use Notification component here as an inline notification?

src/components/Doc/ComponentDoc.js
```diff
@@ -49,6 +50,22 @@ export const ComponentDoc = ({
...
+          <Box
+            alignContent="center"
+            direction="row"
+            background="light-3"
+            pad="xsmall"
+            round="xsmall"
+          >
+            <Text alignSelf="center">
+              <b>New!</b> This component is available on the{' '}
+              <Anchor href="/stable">stable branch</Anchor>
```
halocline, line 64:
```suggestion
              <Anchor href="/stable">stable branch</Anchor>.
```

src/components/Doc/ComponentDoc.js
```diff
@@ -49,6 +50,22 @@ export const ComponentDoc = ({
...
+        <Box pad={{ bottom: 'medium' }}>
+          <Box
+            alignContent="center"
+            direction="row"
+            background="light-3"
+            pad="xsmall"
+            round="xsmall"
+          >
+            <Text alignSelf="center">
+              <b>New!</b> This component is available on the{' '}
```
halocline, line 63: Playing with some messaging:

"New! This component is in beta and available on the stable branch. Test it out and let us know what you think."

Not sure what the plans are for the content on the /stable page, but assuming it explains what the stable branch is, why it exists (i.e. its purpose), and then how to consume it. If so, it can do the heavy lifting explaining how it is used as a staging ground and new stable components are included in the next release.

src/components/Doc/ComponentDoc.js
```diff
@@ -49,6 +50,22 @@ export const ComponentDoc = ({
         <title>{pageTitle || name}</title>
         <meta name="description" content={description} />
       </Helmet>
+      {stable && (
+        <Box pad={{ bottom: 'medium' }}>
+          <Box
+            alignContent="center"
+            direction="row"
+            background="light-3"
+            pad="xsmall"
```
halocline, line 59: Text seems like it doesn't have enough room to breathe and could benefit from additional padding. Using Notification might help here.

halocline (commented):
I'm liking this idea, thanks for pursuing!

I like Mike's comment about how might we make this a bit more dynamic.

Lastly, would it be valuable to have a similar treatment on the /components index page? There could be a dynamic section at the top, ahead of layouts where new components are highlighted. If no new, then the section doesn't appear. Alternatively, could the "tile" for the new component be tagged with a "New" banner or other visual treatment?

src/components/Doc/Property.js
```diff
@@ -27,9 +27,25 @@ export const Property = ({ name, first, children }) => {
...
+            <Box
+              direction="row"
+              gap="xsmall"
+              background="light-2"
+              round="medium"
+              pad={{ vertical: 'xsmall', horizontal: 'small' }}
+            >
+              <CircleInformation />
+              <Text alignSelf="center" size="small">
+                New! available on <Anchor href="/stable">stable branch</Anchor>
```
halocline, line 44:
```suggestion
                New! Available on <Anchor href="/stable">stable branch</Anchor>.
```

src/components/Doc/ComponentDoc.js
jcfilben: Changed this to use the notification component in latest commit

src/components/Doc/ComponentDoc.js
jcfilben: Switched to use Notification component

src/components/Doc/ComponentDoc.js
jcfilben: Fixed

src/components/Doc/ComponentDoc.js
jcfilben: I like this messaging, updated to say, "New! <component name> is in beta, test it out and let us know what you think. This component is available on the stable branch."

The stable page on the site will have the same info as the stable wiki page with maybe a few tweaks https://github.com/grommet/grommet/wiki/What-is-grommet-stable-and-how-to-use-it%3F

src/components/Doc/Property.js
jcfilben: Fixed

jcfilben:
> Would it be possible to pass a version number instead? It would be interesting if you could instead specify what version we expect the ability to be in then the site could do both a 'new in stable' or a 'new in version X' (and only show that if the current version of grommet is that version.) It might help avoid having to go find and clean up old ones with each grommet release.

> I like Mike's comment about how might we make this a bit more dynamic.
>
> Lastly, would it be valuable to have a similar treatment on the /components index page? There could be a dynamic section at the top, ahead of layouts where new components are highlighted. If no new, then the section doesn't appear. Alternatively, could the "tile" for the new component be tagged with a "New" banner or other visual treatment?

Good suggestions! Brittany and I chatted and are thinking these enhancements could come later down the line and for the context of this PR we will just move forward with the stable notifications that we have. I like these ideas though and I will open a separate ticket for this so we don't lose track of them.

src/components/Doc/ComponentDoc.js
```diff
@@ -49,6 +50,21 @@ export const ComponentDoc = ({
...
         <meta name="description" content={description} />
       </Helmet>
+      {stable && (
+        <Notification
+          background="light-3"
+          status="info"
+          global
+          actions={[
+            {
+              label: 'stable branch.',
```
halocline, line 60: I prefer to limit the anchor to the actual subject being linked to, and have the punctuation live outside. Soft opinion.

Out of curiosity, is there an advantage to using the `actions` prop for the link vs. inline with the message? Mmmmm.... seeing that `message` only accepts a string currently -- supporting a node may be a future enhancement. So, this is okay for now.

src/components/Doc/ComponentDoc.js
```diff
@@ -49,6 +50,21 @@ export const ComponentDoc = ({
...
+              href: '/stablebranch',
+              target: '_blank',
+            },
+          ]}
+          message={`New! ${name} is in beta, test it out and let us know what you think. This component is available on the`}
```
halocline, line 65: The text is feeling crowded to be by the containing Box. Playing around with padding to give it more room to breathe, then adding some corner rounding to take the hard edges off and be more harmonious with the button style and interactive code container.

Also, I threw in a bit more spacing between the notification and the page title -- the title is getting squeezed. Open to whether the spacing is applied as a gap on a bounding Box or as margin on the Notification.

```suggestion
          message={`New! ${name} is in beta, test it out and let us know what you think. This component is available on the`}
          round="xsmall"
          pad={{ horizontal="medium", vertical: "small" }}
```

**Before**
Screen Shot 2022-09-12 at 9 30 11 AM

**After**
Screen Shot 2022-09-12 at 9 29 58 AM

src/components/Doc/ComponentDoc.js
```diff
@@ -19,6 +19,7 @@ export const ComponentDoc = ({
   name,
   pageTitle,
   designSystemLink,
+  stable,
```
halocline, line 22: When we're ready, swapping this out with a prop name like `version` or `grommetVersion` could provide flexibility to promote new components in the current release and swap out messaging. But agree with your [comment](https://github.com/grommet/grommet-site/pull/416#issuecomment-1242353846) that we can treat as a future enhancement. Just planting seeds.

halocline (approved):
Looking good. 

Dropped a couple styling suggestions for consideration. Adopt as you see fit. 
Also, made note of a couple future enhancements for which we might want issues created.

Otherwise, this looks good to me.

src/components/Doc/ComponentDoc.js
ericsoderberghp: I'd recommend addressing any styling issues with the base grommet theme, if we're feeling that all notifications would benefit, especially with the text padding aspect here, and probably the rounding.

britt6612: Looks alot better thanks @halocline for your changes

src/components/Doc/ComponentDoc.js
britt6612: I believe when we added the `actions` and `messages` we wanted to limit users and how they can pass something to `messages` but in grommet world we should expand this to support a node

jcfilben: Yes I agree on having the punctuation outside of the anchor. The way 
Notification is built the actions is always placed at the end so I wasn't able to put the punctuation after the anchor. As you noted `message` only accepts a string I can open an issue on grommet for an enhancement that would let us pass a node.

jcfilben: Issue created: https://github.com/grommet/grommet/issues/6320

src/components/Doc/ComponentDoc.js
jcfilben: Changes look good! 

> I'd recommend addressing any styling issues with the base grommet theme, if we're feeling that all notifications would benefit, especially with the text padding aspect here, and probably the rounding.

Wouldn't a change like this cause backwards compatibility issues with the theme?

src/components/Doc/ComponentDoc.js
jcfilben: Good note
Issue created: https://github.com/grommet/grommet-site/issues/417

src/components/Doc/ComponentDoc.js
ericsoderberghp: If we feel that the "grommet" theme seems a bit "broken", we can change it. It would be the same if did something like change a color to improve accessibility.

jcfilben: I pushed up changes to add a little more spacing between the notification and page title. For the other changes, rounding and extra padding within the notification would it make sense to open up an issue on grommet for this so we can explore if we want to make this change in the theme?

ericsoderberghp: sure, or just go ahead and change the theme

halocline: Agree, make change in Grommet theme. New functionality for Notification introduced, Grommet theme had not yet embraced / adopted that new functionality until now. Changes to grommet theme would be justified.

jcfilben: Sounds good! I will do that

ericsoderberghp (approved):

PR #424, opened 2022-10-04
Restructured Color page

ericsoderberghp:
Show all the colors available, grouped by names we know, but allowing for theme specific color names that are just used via indirect reference.


jcfilben (approved):
Looks good!

PR #439, opened 2022-11-20
Skeleton doc

MikeKingdom:
Added documentation for Skeleton related properties on Box, Text, etc as well as the Skeleton component itself.


src/screens/Text.js
```diff
@@ -132,6 +132,17 @@ const TextPage = () => (
           </PropertyValue>
         </Property>
 
+        <Property name="skeleton">
+          <Description>
+            Any Skeleton properties which will be applied to the Text when in
+            skeleton mode. For the Text to be in skeleton mode, use the
```
taysea, line 138: Should we include this "For the [ component ] to be in skeleton mode" sentence on button / heading as well?

src/screens/Box.js
```diff
@@ -561,6 +561,42 @@ const BoxPage = () => (
...
 
+        <Property name="skeleton">
+          <Description>
+            Make the contents of the box 'skeletonized'. When this property is
+            supplied, any components within the Box that are skeleton-aware will
+            appear as gray boxes instead of their usual text or contents. This
+            is intended as a placeholder representation instead of the real
+            contents when it is still being loaded asynchronously. Some
+            skeleton-aware components are Box (when a border or background is
+            supplied), Text, Heading and Paragraph.
```
taysea, line 572: Paragraph is mentioned here but I don't see skeleton prop Paragraph documentation in this PR.

MikeKingdom: Paragraph doesn't have any skeleton-related props to document. It does react to being in a SkeletonContext however.

src/screens/Text.js
MikeKingdom: good point

taysea (approved):
Thanks!

taysea:
Just need to resolve the yarn.lock conflict before merging.

PR #451, opened 2023-02-09
Added Data related components

ericsoderberghp:


src/screens/Data.js
```diff
@@ -0,0 +1,274 @@
...
+          badge:
+            'https://cdn-images-1.medium.com/fit/c/120/120/1*TD1P0HtIH9zF0UEH28zYtw.png',
+          label: 'Storybook',
+        },
+        {
+          url: 'https://github.com/grommet/grommet/tree/master/src/js/components/Data',
+          label: 'Github',
+        },
+      ]}
+      description="Container and context for data driven content"
```
taysea, line 126:
```suggestion
      description="Container and context provider for data driven content"
```
is this too technical?

src/screens/Data.js
```diff
@@ -0,0 +1,274 @@
...
+  toolbar
+>
+  <DataTable />
+</Data>`}
+    >
+      <Properties>
+        <Property name="data">
+          <Description>Array of data objects.</Description>
+          <PropertyValue type="array">
+            <Example>{`[{}]`}</Example>
```
taysea, line 139: Would it be helpful to have mock data here with a few items as opposed to empty object?

src/screens/Data.js
```diff
@@ -0,0 +1,274 @@
...
+        <Property name="defaultView">
+          <Description>The default view of the data.</Description>
+          <PropertyValue type="object">
+            <Example>{viewObjectExample}</Example>
+          </PropertyValue>
+        </Property>
+
+        <Property name="filteredTotal">
+          <Description>
+            The total number of objects that exist in the filtered data set.
```
taysea, line 152: This property is new from the first pass, I don't understand why we have to provide filteredTotal and total. How does that effect the behavior?

src/screens/Data.js
```diff
@@ -0,0 +1,274 @@
...
+          </Description>
+          <PropertyValue type="number">
+            <Example>270</Example>
+          </PropertyValue>
+        </Property>
+
+        <Property name="id">
+          <Description>
+            A unique identifier for the Data. This is required if more than one
+            Data is rendered.
```
taysea, line 165:
```suggestion
            Data is rendered on the page.
```

src/screens/Data.js
```diff
@@ -0,0 +1,274 @@
...
+            <Example>{messagesObjectExample}</Example>
+          </PropertyValue>
+        </Property>
+
+        <Property name="onView">
+          <Description>
+            Function that will be called when the user changes the view. This
+            includes, searching, filtering, sorting, and paging. It will be
+            called with an object describing the current view. See the 'view'
+            property for an example of the object. Callers can combing 'view'
```
taysea, line 186:
```suggestion
            property for an example of the object. Callers can combine 'view'
```

src/screens/Data.js
```diff
@@ -0,0 +1,274 @@
...
+            and 'onView' to use Data as a controlled component.
+          </Description>
+          <PropertyValue type="function">
+            <Example>{`(view) => {}`}</Example>
+          </PropertyValue>
+        </Property>
+
+        <Property name="properties">
+          <Description>
+            This describes the objects found in 'data', sort of a schema.
```
taysea, line 196:
```suggestion
            This describes the objects found in 'data', sort of a schema. Using this property overrides the default filters. It allows caller to specify how property label renders and which properties should be filterable.
```

src/screens/PageHeader.js
```diff
@@ -394,7 +394,7 @@ export const PageHeaderItem = ({ name, path }) => (
     <PageHeader
       title="Page Title"
       subtitle="A subtitle for the page."
-      parent={<Anchor label="Parent Page" />}
+      parent={<Anchor as="span" label="Parent Page" />}
```
taysea, line 397: Why is this added?

src/structure.js
```diff
@@ -63,6 +63,23 @@ export const structure = {
...
+      name: 'Data',
+      components: [
+        // 'Cards',
+        'Data',
+        'DataFilter',
+        'DataFilters',
+        'DataSearch',
+        'DataSort',
+        'DataSummary',
+        'DataTable',
```
taysea, line 76: Nice, I like encouraging people to use DataTable, etc. by grouping this way.

src/screens/Data.js
ericsoderberghp: done

src/screens/PageHeader.js
ericsoderberghp: The browser complains about `<a><a /></a>`, since the PageHeaderItem is rendered inside an anchor. This prevents that console error message.

src/screens/Data.js
ericsoderberghp: I've updated the description to:
> The total number of objects that exist in the data set AFTER any filtering has been applied. This should be used when retrieving data from a paged API where the number of objects in 'data' is only one page and the filtered set of data is larger than a page.
Does that help?

taysea (approved):

PR #458, opened 2023-02-24
Remove button.icon.size

jcfilben:
This change was reverted on grommet so we should remove it from the docs

https://github.com/grommet/grommet/pull/6651


taysea (approved):
Thank you!

PR #467, opened 2023-04-05
Added missing labelKey documentation

coltonw:
Yarn is so woefully out of date here that Volta won't even run it, so I couldn't test this locally but it is a copy paste of the documentation from the regular Select so it should be solid.


jcfilben (approved):
Thanks for adding this!

ericsoderberghp (approved):

PR #472, opened 2023-04-18
Added DataTableGroupBy

ericsoderberghp:


src/screens/DataTableGroupBy.js
```diff
@@ -0,0 +1,63 @@
...
+  Description,
+  Example,
+} from '../components/Doc';
+
+const DataTableGroupByPage = () => (
+  <Page>
+    <ComponentDoc
+      name="DataTableGroupBy"
+      stable
+      availableAt={[
```
jcfilben, line 19: Can we add a storybook link here? https://storybook.grommet.io/?path=/story/data-datatablegroupby-simple--simple

jcfilben (commented):
Looks like `yarn.lock` needs updated again

src/screens/DataTableGroupBy.js
ericsoderberghp: good catch, done

jcfilben (approved):
Looks good, just need to update yarn.lock

ericsoderberghp:
> Looks good, just need to update yarn.lock

I think the issue is that the NODE_VERSION in Netlify needs to be updated. Done, rerunning ...

Issue #480, opened 2023-06-05
Document `cssGap` option in Box gap


jcfilben:
We should make note of the `<Grommet options={{ box: { cssGap: true } }} />` option in the Box `gap` documentation so that it is easier to find and will bring increased awareness to this option.

PR #481, opened 2023-06-05
Document Grommet's box cssGap option on Box `gap` prop

jcfilben:
You can change the behavior of Box's gap by setting `options={{box: { cssGap: true } }}` on the Grommet container. This is documented on the Grommet options prop but not on the Box gap prop. This PR makes note of this on the Box gap prop description.

Closes https://github.com/grommet/grommet-site/issues/480

### DIFF
--- a/src/screens/Box.js
+++ b/src/screens/Box.js
@@ -374,10 +374,12 @@ const BoxPage = () => (
 
         <Property name="gap">
           <Description>
-            The amount of spacing between child elements. This should not be
+            {`The amount of spacing between child elements. This should not be
             used in conjunction with 'wrap' as the gap elements will not wrap
             gracefully. If a child is a Fragment, Box will not add a gap between
-            the children of the Fragment.
+            the children of the Fragment. To change the behavior of gap to align
+            with the CSS gap attribute, set options={{ box: { cssGap: true } }} 
+            on the Grommet container.`}
           </Description>
           <PropertyValue type="string">
             <Description>
--- a/src/screens/Box.js
+++ b/src/screens/Box.js
@@ -377,9 +377,10 @@ const BoxPage = () => (
             {`The amount of spacing between child elements. This should not be
             used in conjunction with 'wrap' as the gap elements will not wrap
             gracefully. If a child is a Fragment, Box will not add a gap between
-            the children of the Fragment. To change the behavior of gap to align
-            with the CSS gap attribute, set options={{ box: { cssGap: true } }} 
-            on the Grommet container.`}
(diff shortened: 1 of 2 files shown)


src/screens/Box.js
```diff
@@ -374,10 +374,12 @@ const BoxPage = () => (
 
         <Property name="gap">
           <Description>
-            The amount of spacing between child elements. This should not be
+            {`The amount of spacing between child elements. This should not be
             used in conjunction with 'wrap' as the gap elements will not wrap
             gracefully. If a child is a Fragment, Box will not add a gap between
-            the children of the Fragment.
+            the children of the Fragment. To change the behavior of gap to align
+            with the CSS gap attribute, set options={{ box: { cssGap: true } }}
```
MikeKingdom, line 381: Saying "align with the CSS gap attribute" sounds odd to me and we should say where to set that option. Maybe say something more like "To use the CSS gap attribute to add space between children of the Box (including children of a Fragment child of the Box), set options={{ box: { cssGap: true } }} on the nearest Grommet parent component.

MikeKingdom (approved):
Looks great and I double checked it actually works that way :)

halocline:
> Looks great and I double checked it actually works that way :)

Is there value to allowing an opt-in at the instance level of a particular Box as well? I can see the value at the Grommet level, but what if I want to gradually convert? I would suspect I'd run into issues by opting in all at once, and desiring some sort of migration path. Thoughts?

MikeKingdom:
> > Looks great and I double checked it actually works that way :)
> 
> Is there value to allowing an opt-in at the instance level of a particular Box as well? I can see the value at the Grommet level, but what if I want to gradually convert? I would suspect I'd run into issues by opting in all at once, and desiring some sort of migration path. Thoughts?

Great question since I was just thinking I'd love to use this but it's tough to in an existing app. I'd vote for adding this as a follow on effort.

jcfilben:
> > Is there value to allowing an opt-in at the instance level of a particular Box as well? I can see the value at the Grommet level, but what if I want to gradually convert? I would suspect I'd run into issues by opting in all at once, and desiring some sort of migration path. Thoughts?
> 
> Great question since I was just thinking I'd love to use this but it's tough to in an existing app. I'd vote for adding this as a follow on effort.

Good idea, I'm supportive of adding that.

jcfilben:
I'm going to merge this PR into the docs for now. Once we have the behavior that allows us to specify cssGap at the Box level we can go back in and adjust the docs.

jcfilben:
Created an issue for adding cssGap opt-in at the Box level: https://github.com/grommet/grommet/issues/6832

PR #483, opened 2023-06-09
add doc for menu item align

britt6612:
Add `menu.item.align` also changed example of `menu.item` to `alignSelf` since Button does not have an `align` prop


src/screens/Menu.js
```diff
@@ -423,6 +423,17 @@ const MenuPage = () => (
...
+        </Property>
+
+        <Property name="menu.item.align">
+          <Description>
+            How to align the contents along the cross axis. Any 'align-items'
+            valid CSS value is accepted, including composed ones such 'first
+            baseline' and 'unsafe start'.
+          </Description>
+          <PropertyValue type="string">
             <Example>{`{ align: 'start' }`}</Example>
```
jcfilben, line 437:
```suggestion
            <Example>'start' </Example>
```
Can we just change this example to a string since we specify the align part in the property name

britt6612: agree

jcfilben (approved):
Looks good, thanks!

PR #485, opened 2023-07-05
Update `bounds` and add `direction in DataChart and Chart

jcfilben:
Update the docs pre grommet release. Added doc changes for https://github.com/grommet/grommet/pull/6799

### DIFF
--- a/src/screens/DataChart.js
+++ b/src/screens/DataChart.js
@@ -355,7 +355,7 @@ const DataChartPage = () => (
           <Description>The orientation of the data.</Description>
           <PropertyValue type="string">
             <Example>"horizontal"</Example>
-            <Example>"vertical"</Example>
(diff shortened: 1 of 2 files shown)


src/screens/DataChart.js
```diff
@@ -350,6 +351,14 @@ const DataChartPage = () => (
           <GenericBool />
         </Property>
 
+        <Property name="direction">
+          <Description>The orientation of the data.</Description>
+          <PropertyValue type="string">
+            <Example>"horizontal"</Example>
+            <Example>"vertical"</Example>
```
taysea, line 358: should this be marked as default value?

jcfilben: Yes, good call. I'll update that

taysea (approved):
LGTM. Thanks for adding.

PR #488, opened 2023-09-01
Upgrade dependencies

taysea:


package.json
```diff
@@ -28,7 +28,7 @@
   },
   "dependencies": {
     "cheerio": "^1.0.0-rc.3",
-    "core-js": "^3.31.1",
+    "core-js": "^3.32.1",
     "grommet": "https://github.com/grommet/grommet/tarball/stable",
     "grommet-icons": "^4.11.0",
     "grommet-styles": "^0.2.0",
```
jcfilben, line 34: Are we still using grommet-styles? I'm not seeing it used anywhere. Would be nice to get rid of it if possible

taysea: Good note. Removed.

jcfilben (approved):
Looks good!

PR #489, opened 2023-10-04
feat: update markdown content example

ho823:
Update markdown example in documentation grommet


jcfilben:
Hi @ho823 can you try running `rm -rf yarn.lock && rm -rf node_modules && yarn cache clean && yarn install`. That should fix the yarn.lock issue

PR #492, opened 2023-11-16
add dataSearch `as` TextInput

britt6612:
Added in the docs that 
```
DataSearch is a [TextInput](http://localhost:8567/textinput). You can customize it using the properties available in TextInput.

```

We have this wording already so just utilized this.
In DataSearch.js in grommet we have 
```
      icon={<Search />}
      type="search"

```
for the TextInput prop however being that this is a DataSearch hence Search the keyword not sure if we should add these in the docs as default open to suggestions

### DIFF
--- a/src/screens/DataSearch.js
+++ b/src/screens/DataSearch.js
@@ -29,6 +29,10 @@ const DataSearchPage = () => (
   <Toolbar><DataSearch /></Toolbar>
   <DataTable />
 </Data>`}
+      isA={{
+        base: 'TextInput',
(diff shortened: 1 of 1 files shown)


taysea:
I'm personally comfortable leaving as is and don't necessarily feel the icon/type are needed right now.

taysea (approved):

jcfilben (approved):
Looks good, I agree with Taylor's assessment on not needing the icon/type in the docs for now

PR #493, opened 2023-11-17
add in description for options to also take checkbox props

britt6612:
DataTableColumns `options` prop also takes any `checkbox` props. Adding this in the description so users can know. One difference is that instead of `value` prop in CheckBox in DataTableColumns it needs a `property` passed in if an array of objects is given.

### DIFF
--- a/src/screens/DataTableColumns.js
+++ b/src/screens/DataTableColumns.js
@@ -46,8 +46,12 @@ const DataTableColumnsPage = () => (
             <Example>{`['id', 'name', 'age']`}</Example>
           </PropertyValue>
           <PropertyValue type="array of objects">
+            <Description>
+              An array of objects of CheckBox props excluding the 'checked'
+              property, use 'property' prop instead of 'checked'.
+            </Description>
             <Example>{`[
-  { label: 'Id', value: 'id' },
+  { label: 'Id', value: 'id', disabled: true },
   { label: 'Name', value: 'name' },
   { label: 'Age', value: 'age' },
 ]`}</Example>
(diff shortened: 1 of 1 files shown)


src/screens/DataTableColumns.js
```diff
@@ -46,8 +46,12 @@ const DataTableColumnsPage = () => (
             <Example>{`['id', 'name', 'age']`}</Example>
           </PropertyValue>
           <PropertyValue type="array of objects">
+            <Description>
+              An array of objects of CheckBox props excluding the 'checked'
```
taysea, line 50: Wording suggestion: "An array of objects, where the object contains CheckBox props. The CheckBox `checked` property is excluded to allow DataTableColumns to manage the state."

@britt6612 Can you clarify what `property` section means? I'm wondering if we should document or exclude.

britt6612: yeah it took me a while to figure it out as well until I looked in the `index.d.ts` and saw 
```
  options: (string | { property: string; label: string })[];
```

britt6612: so Ill add this property into the docs as well

britt6612: also not sure about that naming `properties` and why it wasn't `value`

taysea: Or if we don't want to support all CheckBox props, but we do want to allow `disabled` specifically, we should be more specific and just add that.

taysea: Based on discussion with @jcfilben, we've decided to document `disabled` for now and add typescript support for it in Grommet. However, since use cases haven't introduced a need for any other CheckBox props yet, we won't support those in Typescript, prop types, or docs.

See: https://github.com/grommet/grommet/pull/7028

taysea: I pushed a wording change to align with this.

taysea (approved):

jcfilben (approved):
Looks good!

britt6612:
Thanks ladies 🎉

PR #494, opened 2023-11-20
Add documentation for DataFilter step

taysea:
Add documentation related to this PR that added `step` to DataFilter range: https://github.com/grommet/grommet/pull/6942

### DIFF
--- a/src/screens/DataFilter.js
+++ b/src/screens/DataFilter.js
@@ -57,7 +57,7 @@ const DataFilterPage = () => (
         <Property name="range">
           <Description>
             The range of numeric values. Should not be used with 'options'. If
-            not specified, the default step is 20.
+            not specified, the default step will be calculated as (max - min) / 20.
           </Description>
           <PropertyValue type="object">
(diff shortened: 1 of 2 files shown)


src/screens/DataFilter.js
```diff
@@ -56,10 +56,11 @@ const DataFilterPage = () => (
 
         <Property name="range">
           <Description>
-            The range of numeric values. Should not be used with 'options'.
+            The range of numeric values. Should not be used with 'options'. If
+            not specified, the default step is 20.
```
jcfilben, line 60: Isn't the default step the total range divided by 20?

taysea: You're right, let me adjust the wording.

taysea, line 60:
```suggestion
            not specified, the default step will be calculated as (max - min) / 20.
```

taysea: Adjusted this wording.

jcfilben (approved):
Looks good!

PR #497, opened 2023-11-27
remove routed button for grommet site

Sulaymon333:
Remove reouted button in favour of Button from grommet v2.
Button component is now in use for grommet-site

### DIFF
--- a/src/components/Search.js
+++ b/src/components/Search.js
@@ -9,6 +9,7 @@ const allSuggestions = structure.sections
   .map((section) => (section.components || []).concat(section.name))
   .concat(structure.externals.map((e) => e.name))
   .reduce((acc, val) => acc.concat(val), [])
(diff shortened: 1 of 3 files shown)


MikeKingdom (commented):
Why remove this? The routed button is there so it uses the router to navigate internal links rather than just an href link which will cause the entire page to reload.

MikeKingdom:
So given Jessica's clarification, the only file that needs to be changed is Examples.js which is the only one using RoutedButton from Grommet. And how that one should be changed depends on what it's trying to provide an Example of. And I actually can't find anywhere Examples.js is used

Sulaymon333:
Yes, Example.js is not used at all, it is not part of the routed pages either, there is no direct way/button to navigate to it. Do we still need to keep it in grommet-site?

PR #500, opened 2023-12-12
Update desc toolbar

britt6612:
change default to be `layer` and no longer drop


jcfilben (approved):
Looks good!

src/screens/Data.js
```diff
@@ -245,7 +245,7 @@ const DataPage = () => (
           <PropertyValue type="boolean">
             <Description>
               Setting it to 'true' will include a Toolbar containing both
-              DataSearch and DataFilters with 'drop'.
+              DataSearch and DataFilters within a 'layer'.
```
taysea, line 248:
```suggestion
              DataSearch and DataFilters with `layer` prop.
```
Wondering how we might specify this is a prop applied

taysea (approved):

PR #508, opened 2024-02-09
Update DataFilters doc to better convey that inline is the default behavior

jcfilben:
In the DataFilters doc show that `layer` and `drop` props are `false` by default. I also changed the example on this page to show DataFilters inline instead of in a drop to give people a better idea of the default behavior.


britt6612 (approved):
Makes more sense

taysea (approved):
Thank you for updating this!

PR #509, opened 2024-02-12
Added additional details to SelectMultiple `onMore` and `onSearch`

jcfilben:


britt6612 (approved):

src/screens/SelectMultiple.js
```diff
@@ -305,7 +305,8 @@ const SelectMultiplePage = () => (
           <Description>
             Function that will be called when the user types in the search
             input. If this property is not provided, no search field will be
-            rendered.
+            rendered. New options can be added to 'options' based on a search
```
taysea, line 308: I'm wondering if this sentence would make sense to someone less familiar with the pattern. is there anything that could be added in the example code snippet that might make this more clear?

taysea: I think that wording is great. Maybe just "search term" as opposed to "search term query".

jcfilben: Sounds good, I'll update and push up changes

taysea (approved):
LGTM.

PR #512, opened 2024-03-05
Add css gap

britt6612:
added behavior that was enhanced here https://github.com/grommet/grommet/pull/7147

### DIFF
--- a/src/screens/Box.js
+++ b/src/screens/Box.js
@@ -321,6 +321,14 @@ const BoxPage = () => (
           </PropertyValue>
         </Property>
 
+        <Property name="cssGap">
+          <Description>
+            Lets users opt into using the css gap property for Box gap instead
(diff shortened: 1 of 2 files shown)


src/screens/Box.js
```diff
@@ -391,6 +400,15 @@ const BoxPage = () => (
             <SizesXsmallXlarge />
             <Example>"any CSS size"</Example>
           </PropertyValue>
+          <PropertyValue type="object">
+            <Description>
+              Passing an object to gap will automatially use the cssGap behavior
```
taysea, line 405: Wording suggestion:

Passing an object to gap will automatically use the cssGap behavior. This is recommended when using `wrap` to avoid extra DOM nodes. Any t-shirt size or specific size in px, em, etc. is supported.

britt6612: updated

PR #513, opened 2024-03-06
Add doc changes pre 2.36.0 release

jcfilben:
Added doc changes for the following PRs: 
RangeInput wheel https://github.com/grommet/grommet/pull/7130
Readonly https://github.com/grommet/grommet/pull/7123

### DIFF
--- a/src/screens/DateInput.js
+++ b/src/screens/DateInput.js
@@ -3,6 +3,7 @@ import { Box, DateInput } from 'grommet';
 import Page from '../components/Page';
 import Item from './Components/Item';
 import {
+  GenericBackground,
   GenericBoolFalse,
   SizesXsmallXlarge,
 } from '../utils/genericPropExamples';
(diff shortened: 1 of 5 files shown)


src/screens/DateInput.js
```diff
@@ -147,6 +148,13 @@ const DateInputPage = () => (
           </PropertyValue>
         </Property>
 
+        <Property name="readOnlyCopy">
+          <Description>
+            Makes the TextInput readOnly and adds a copy to clipboard button.
```
britt6612, line 153: were we going to mention that they can use the `readOnly` alone or since that's a native behavior were not mentioning it? I don't think its own section but maybe within this description but not strong opinion if we want to leave off

jcfilben: Yeah I didn't document it because it is a native prop. But maybe I could add something in the `readOnlyCopy` description

jcfilben: Just pushed up changes. I just mentioned the `readOnly` prop more explicitly within the `readOnlyCopy` description so hopefully that helps

src/screens/DateInput.js
```diff
@@ -147,6 +148,15 @@ const DateInputPage = () => (
...
```
taysea, line 153:
```suggestion
            Makes the DateInput readOnly and adds a copy to clipboard button.
```

britt6612 (approved):

PR #516, opened 2024-04-18
add docs for new enhancements for April release

britt6612:
Add Skeleton `extend`
Add `round` to responsive doc
Add `pinned` behavior

### DIFF
--- a/src/screens/DataTableColumns.js
+++ b/src/screens/DataTableColumns.js
@@ -49,7 +49,7 @@ const DataTableColumnsPage = () => (
               An array of objects that can include a `label` which will render
               as the option label, `property` which should be a data property,
                `disabled` which will determine if the option can be
-              checked/unchecked. `pinned` which can be used to specify items
+              checked/unchecked, and `pinned` which can be used to specify items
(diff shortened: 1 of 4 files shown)


taysea (approved):
Likely will need a yarn.lock update before we merge

src/screens/DataTableColumns.js
```diff
@@ -49,11 +49,12 @@ const DataTableColumnsPage = () => (
               An array of objects that can include a `label` which will render
               as the option label, `property` which should be a data property,
               and `disabled` which will determine if the option can be
```
taysea, line 51:
```suggestion
               `disabled` which will determine if the option can be
```

src/screens/DataTableColumns.js
```diff
@@ -49,11 +49,12 @@ const DataTableColumnsPage = () => (
...
-              checked/unchecked.
+              checked/unchecked. `pinned` which can be used to specify items
```
taysea, line 52:
```suggestion
              checked/unchecked, and`pinned` which can be used to specify items
```

PR #519, opened 2024-04-26
fix link

britt6612:
Did not have the link for storybook for ToggleGroup

### DIFF
--- a/src/screens/ToggleGroup.js
+++ b/src/screens/ToggleGroup.js
@@ -20,9 +20,9 @@ const ToggleGroupPage = () => (
       name="ToggleGroup"
       availableAt={[
         {
-          url: '',
+          url: 'https://storybook.grommet.io/?path=/story/controls-togglegroup-simple--simple',
           badge:
-            'https://storybook.grommet.io/?path=/story/controls-togglegroup-simple--simple',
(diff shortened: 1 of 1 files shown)


halocline (approved):

src/screens/ToggleGroup.js
```diff
@@ -20,9 +20,9 @@ const ToggleGroupPage = () => (
       name="ToggleGroup"
       availableAt={[
         {
-          url: '',
+          url: 'https://storybook.grommet.io/?path=/story/controls-togglegroup-simple--simple',
           badge:
-            'https://storybook.grommet.io/?path=/story/controls-togglegroup-simple--simple',
+            'https://cdn-images-1.medium.com/fit/c/120/120/1*TD1P0HtIH9zF0UEH28zYtw.png',
```
halocline, line 25: Where does this badge appear?

jcfilben: I think we used to have icons next to the storybook and codesandbox links, but looks like those are not showing up. I don't remember if we intentionally removed them or not

jcfilben: If I had to guess I would say we intentionally removed them, but then I'm not sure why we kept the Github icon

britt6612: yeah I looked in `componentDoc` and saw `badge` was never being used

britt6612: removed!

PR #533, opened 2025-01-22
add calendar level prop

britt6612:
updated yarn.lock & add `level` prop to Calendar


britt6612:
yarn lock updates the new status colors as well

jcfilben (approved):
Looks good!

PR #539, opened 2025-04-01
Clarify when to use primaryKey instead or columns.primary

taysea:
Based on Slack discussion, clarify when to use `primaryKey` instead of `primary` in columns object. In cases where the primary column gets hidden (for example with DataTableColumns), this can cause data to render incorrectly if `primaryKey` is not defined.

### DIFF
--- a/src/screens/DataTable.js
+++ b/src/screens/DataTable.js
@@ -199,10 +199,12 @@ const DataTablePage = () => (
               property should be used as the unique identifier, which gives the
               cell 'row' scope for accessibility. If 'primary' is not used for
               any column, and 'primaryKey' isn't specified either, then the
-              first column will be used. 'pin' indicates that this column should
-              not scroll out of view to the left when the table is scrolled
-              horizontally. 'plain' = true indicates that the body cells in the
-              column will not apply pad.
+              first column will be used. If it's possible the primary column may
+              be removed from the UI (i.e., when hidden using DataTableColumns),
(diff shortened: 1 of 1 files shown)


jcfilben (approved):

britt6612 (approved):

src/screens/DataTable.js
```diff
@@ -199,10 +199,12 @@ const DataTablePage = () => (
               property should be used as the unique identifier, which gives the
               cell 'row' scope for accessibility. If 'primary' is not used for
               any column, and 'primaryKey' isn't specified either, then the
-              first column will be used. 'pin' indicates that this column should
-              not scroll out of view to the left when the table is scrolled
-              horizontally. 'plain' = true indicates that the body cells in the
-              column will not apply pad.
+              first column will be used. If it's possible the primary column may
+              be removed from the UI (i.e., when hidden using DataTableColumns),
+              use 'primaryKey' instead to ensure data renders correctly. 'pin'
```
halocline, line 204: hmmm.... This seems to be difficult to discover insider knowledge...

Wondering if we might be able to honor `columns.myProperty.primary` more elegantly?

LiveFreeAndRoam (commented):
For our case, the proposed changes are clear to me.

I agree with @halocline comment to.

src/screens/DataTable.js
taysea: File a grommet ticket to track: https://github.com/grommet/grommet/issues/7580

PR #541, opened 2025-04-24
Update pagination message prop

Sulaymon333:
- fix/update pagination message prop


jcfilben (commented):
Can we also add this to the Grommet component page `message` prop

Sulaymon333:
I updated the grommet message prop for pagination

PR #566, opened 2026-08-18
Add docs for DateTimeInput

jcfilben:
Adds docs for new DateTimeInput component


src/screens/DateTimeInput.js
```diff
@@ -0,0 +1,300 @@
...
+  <Page>
+    <ComponentDoc
+      name="DateTimeInput"
+      availableAt={[
+        {
+          url: 'https://github.com/grommet/grommet/tree/master/src/js/components/DateTimeInput',
+          label: 'GitHub',
+        },
+      ]}
+      description="a control to input a time and date value"
```
britt6612, line 26:
```suggestion
      stable
      description="a control to input a time and date value"
```

src/screens/DateTimeInput.js
```diff
@@ -0,0 +1,300 @@
...
+
+        <Property name="dateTimeInput.drop.gap">
+          <Description>The gap between elements inside the drop.</Description>
+          <PropertyValue type="string">
+            <Example defaultValue>"small"</Example>
+          </PropertyValue>
+        </Property>
+
+        <Property name="dateTimeInput.drop.border.color">
+          <Description>The border color for the drop container.</Description>
```
britt6612: The color of the divider between the Calendar and TimeInput inside the
    DateTimeInput drop.


wonder if that would be clear

src/screens/DateTimeInput.js
```diff
@@ -0,0 +1,300 @@
...
+          <PropertyValue type="string">
+            <Example defaultValue>"border"</Example>
+          </PropertyValue>
+        </Property>
+
+        <Property name="dateTimeInput.drop.border.size">
+          <Description>The border size for the drop container.</Description>
```
britt6612: The border size of the divider between the Calendar and TimeInput inside
    the DateTimeInput drop.

src/screens/DateTimeInput.js
jcfilben: Good catch!

src/screens/Grommet.js
```diff
@@ -349,6 +349,21 @@ const GrommetPage = () => (
     dataView: {
       label: "string"
     },
+    dateTimeInput: {
```
britt6612, line 352: I dont think we have a specific pattern here since some say what we have in the default json file I noticed in Wizard you put the default string should we do the same here?

jcfilben: Yeah we aren't very consistent, we should determine a pattern and align to that. I'll leave this for now until we determine a direction

src/screens/DateTimeInput.js
jcfilben: Changed to this in latest commit

src/screens/DateTimeInput.js
jcfilben: Changed to this in latest commit

src/screens/Grommet.js
jcfilben: Ticket created: https://github.com/grommet/grommet-site/issues/567

britt6612: great thank you

src/screens/DateTimeInput.js
```diff
@@ -0,0 +1,307 @@
...
+      </ThemeDoc>
+    </ComponentDoc>
+  </Page>
+);
+
+export default DateTimeInputPage;
+
+export const DateTimeInputItem = ({ name, path }) => (
+  <Item name={name} path={path} center pad={{ horizontal: 'xlarge' }}>
+    <DateTimeInput disabled />
```
britt6612: Image


can we add a defaultValue here looks the same as DateInput

jcfilben: Added defaultValue

britt6612 (approved):

PR #569, opened 2026-08-18
Add beta flag that won't display the stable branch link

MikeKingdom:
This adds a `beta` flag that can be used instead of the `stable` flag and changed the DateTimeInput, TimeInput, Wizard and Stepper to use that flag instead of stable since they are available in a versioned release.

image

The stable version of the message is still shown if the `stable` flag is used.


jcfilben (approved):
Looks good!

britt6612 (approved):

PR #572, opened 2026-08-27
add fix to check-drift workflow

britt6612:
The action for the check-drift was failing since sometimes grommet stable branch with the hash changes. Now it will do the new install which will refresh the grommet lock all in one pass so shouldn't be any more stale hash.


jcfilben (approved):
Thanks for fixing!

MikeKingdom (approved):

